PR #44, opened 2015-03-11
Change magnitude to signed_magnitude_10.

chitang:


common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -64,7 +64,7 @@ object SqlFunctions {
     // http://beta.dev.socrata.com/docs/datatypes/numeric.html
     UnaryPlus -> formatCall("%s"),
     UnaryMinus -> formatCall("-%s") _,
-    Magnitude -> formatCall("floor(log(%s, abs(%s) * %s + 1))", Some(Seq(1,0,1,0))),
```
znep: So we had the open question of if we can make 0 map to 0 with a case statement without impacting performance.  Did you try that?

znep: Ok, @chitang pointed out that sign(0) is 0 so our concern about 0 is misplaced.

soql-server-pg/src/test/resources/fixtures/select-signed_magnitude_10.json
```diff
@@ -0,0 +1,13 @@
...
+  { "name" : "Chili", "line_length" : -1, "signed_magnitude_10_line_length" : -1 },
+  { "name" : "Karma", "line_length" : 412, "signed_magnitude_10_line_length" : 3 },
+  { "name" : "Karma SP", "line_length" : 412, "signed_magnitude_10_line_length" : 3 },
+  { "name" : "Mescal", "line_length" : 295, "signed_magnitude_10_line_length" : 3 },
+  { "name" : "Rush", "line_length" : -1, "signed_magnitude_10_line_length" : -1 },
+  { "name" : "Sprint Evo", "line_length" : null, "signed_magnitude_10_line_length" : null },
+  { "name" : "Tequila", "line_length" : -1, "signed_magnitude_10_line_length" : -1 },
+  { "name" : "Vista II SP", "line_length" : 250, "signed_magnitude_10_line_length" : 3 },
+  { "name" : "Zefira", "line_length" : 285, "signed_magnitude_10_line_length" : 3 }
+]
```
znep: So just to cover all the cases, can we stick a 0 in here as a testcase?  And can we test some double values?

znep:
LGTM other than adding a few little test cases.

sokrahta:
cool, looks good

PR #47, opened 2015-03-27
Docker support for pg secondary watcher and soql-server.

znep:
A few TODOMSs left to finish up.


znep:
This is ready for review and merge.

project/Dependencies.scala
```diff
@@ -15,6 +15,7 @@ object Dependencies {
     val scalaTest = "2.1.0-RC2"
     val simpleArm = "1.1.10"
     val slf4j = "1.7.5"
+    val socrataHttpServer = "3.1.0"
```
velvia: You don't need this.  socrataHttpCuratorBroker already pulls in 3.3.0, and you are pulling in a lower version.

soql-server-pg/docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM socrata/java
+
+# TODO expose JMX
+EXPOSE 6090 6091
```
velvia: What is the second port for?

soql-server-pg/docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+ENV SERVER_ROOT /srv/soql-server-pg
+ENV SERVER_ARTIFACT soql-server-pg-assembly.jar
+ENV SERVER_CONFIG soql-server-pg.conf
+
+# defaults
+ENV ENABLE_GRAPHITE false
+ENV GRAPHITE_HOST 0.0.0.0
+ENV GRAPHITE_PORT 0
+ENV JAVA_XMX 512m
```
velvia: 512m seems like a really low default.

soql-server-pg/docker/ship.d/run
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -ev
+
+/bin/env_parse ${SERVER_CONFIG}.j2
+
+exec su socrata -c '/usr/bin/java \
+    -Xmx${JAVA_XMX} \
+    -Xms${JAVA_XMX} \
+    -Dconfig.file=${SERVER_CONFIG} \
```
velvia: Do we need the absolute path here again?

soql-server-pg/docker/Dockerfile
znep: liveness-check.port = 6091

soql-server-pg/docker/Dockerfile
znep: The intent is it is a default that someone saying "oh, I just want to try running this" can reasonably use.  eg. developer running locally.  If 512 megs isn't enough for that then we should certainly bump it, but it feels like it should be reasonable.

The "this makes sense for a production environment" number is higher, but is a more specific scenario so I don't think it deserves to be the default.

soql-server-pg/docker/ship.d/run
znep: A good question.  I really would prefer to not specify the path under the DRY principle: we said what our context is once (by definition, that is what WORKDIR does), so why repeat it.

I'm confused though because it seems to work fine here and in almost every other project but didn't in the secondary-watcher.  Or I was just confused there.

I'd prefer to stick with not specifying it in general where it works, and go back and figure out why it didn't seem to work in secondary-watcher when we have time.

velvia:
@znep did you leave a build reference to `socrataHttpServer` somewhere?

build failure:
```
0.13/classes...
[error] /scratch/jenkins/workspace/soql-postgres-adapter/project/SoqlServerPG.scala:17: not found: value socrataHttpServer
[error]     socrataHttpServer
[error]     ^
[error] one error found
[error] (compile:compile) Compilation failed
Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? Build step 'Build using sbt' changed build result to FAILURE
```

znep:
@velvia oops, fixed.

PR #52, opened 2015-04-05
CORE-4794 Rollups are not created on resync.

chitang:


store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -420,6 +420,11 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
     }
     sLoader.optimize(truthSchema.values)
 
+    if (truthCopyInfo.lifecycleStage == com.socrata.datacoordinator.truth.metadata.LifecycleStage.Unpublished &&
```
znep: We can use the TruthLifecycleStage alias here instead.

znep:
Awesome, thanks @chitang.  Good to merge modulo the one nit on line 423.

I don't like this codepath and think the real fix is a bigger refactor of that whole method to have fewer codepaths where things can go wrong in different ways.  In particular, the match block at https://github.com/socrata/soql-postgres-adapter/blob/rollup_on_resync/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala#L359-L394 really should be refactored so we only create copies, etc. in one place.  Instead of a different code path if it exists or not, make it always try to remove an existing copy, then have one path for creation regardless.

But that is probably too big and risky of a thing to try to do right now.

PR #53, opened 2015-04-07
Use newer coordinatorlib so tests pass

velvia:
Also, create a log4j config in src/test/resources so no more huge noisy logs during unit tests - they all get logged to target/unit-tests.log


soql-server-pg/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,11 @@
+# Set everything to be logged to the file target/unit-tests.log
```
znep: Hrm.  We may need to figure out some different way to deal with different development tools here.  For example, when I run tests in IntelliJ then I really want messages to go to stdout because IntelliJ automatically groups them under the right tests so I can easily see the logs associated with a failing test.

velvia: Could we do something like use environment vars to point at a different
log4j config, or an SBT option that only gets picked up outside of
IntelliJ?  (Not sure what IntelliJ picks up or doesn't, since now it also
reads files)

On Mon, Apr 6, 2015 at 9:46 PM, Marc Slemko <notifications@github.com>
wrote:

> In soql-server-pg/src/test/resources/log4j.properties
> <https://github.com/socrata/soql-postgres-adapter/pull/53#discussion_r27851377>
> :
>
> > @@ -0,0 +1,11 @@
> > +# Set everything to be logged to the file target/unit-tests.log
>
> Hrm. We may need to figure out some different way to deal with different
> development tools here. For example, when I run tests in IntelliJ then I
> really want messages to go to stdout because IntelliJ automatically groups
> them under the right tests so I can easily see the logs associated with a
> failing test.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/socrata/soql-postgres-adapter/pull/53/files#r27851377>
> .
>



-- 
Evan Chan   |   Principal Engineer, Socrata     |    @evanfchan
*Socrata <http://www.socrata.com/> *| The Data Experience Company™
*Socrata isn't about worldly success, but about changing the world
successfully. This is why I'm excited to be here.*
*What does Socrata do? <https://www.youtube.com/watch?v=yyz86MlkriQ> *(watch
our 2 min video)

znep: It doesn't look like there is an easy way to automatically detect running inside intellij, but you can certainly set a system property explicitly.  The "goal" is for intellij to pick up everything that sbt does.

We could totally have an env variable though, just have to pick the right default.  I agree that just running all the tests in sbt + logging to stdout isn't ideal, but it also isn't ideal to hide the log output and make it harder to associate with a given test when often the only indication of the real cause of failure is in the log...

velvia: @znep ScalaTest does change the thread name to the name of the test, so it's pretty easy to grep out the name of the test.   In fact maybe that's what IntelliJ is doing.   And when you run a lot of tests together, the log output often gets jumbled anyways, meaning you have to grep no matter where it's going.

Anyways I can't seem to use JVM flags to have log4j redirect its output, except for using src/test/resources.  Maybe I'll just keep a local file there and not check it in.

velvia:
@znep there you go.

PR #56, opened 2015-04-22
CORE-4677 - Specify encoding when writing CJSON response

urmilan:


soql-server-pg/src/main/scala/com/socrata/pg/server/CJSONWriter.scala
```diff
@@ -41,6 +41,7 @@ object CJSONWriter {
     Logger(LoggerFactory getLogger getClass.getName)
 
   val dateTimeFormat = ISODateTimeFormat.dateTime
+  val encodingName = scala.io.Codec.UTF8.name
```
kaidamasaki: I might use something more specific than `encodingName`.

kaidamasaki:
Other than the variable name LGTM.

soql-server-pg/src/main/scala/com/socrata/pg/server/CJSONWriter.scala
urmilan: Fixed in 0a2eaab

PR #57, opened 2015-04-23
Text->Money conversion to support queries like 'select * where price > '1.50'

urmilan:
Addresses CORE-4777


znep:
So I already merged this ... but do we also want a number to money conversion?

urmilan:
Was wondering that myself. Giacomo asked for text to number so I implemented that first, was gonna get a hold of him and do that in a separate PR if it's required.

znep:
Just seems to make sense from a consistency standpoint even if it isn't an
immediate frontend requirement...

PR #58, opened 2015-04-24
Number->money conversion

urmilan:


urmilan:
@chitang Addressed CR comments in d4ca0d1 and 04a7200

PR #59, opened 2015-04-25
case function

chitang:
It is here.  Why not include another function?


urmilan:
:+1: LGTM

kaidamasaki:
I aborted this build accidentally.  Ignore the failure (though, I wouldn't assume it was a success, either.)

PR #65, opened 2015-05-07
Drop working copy on a dataset with rollup will cause replication fail.

chitang:
https://socrata.atlassian.net/browse/CORE-5429


chitang:
There is another problem about dataVersion.  Working on it.

PR #66, opened 2015-05-11
Align configuration root path so that migration can run.

chitang:


velvia:
Looks fine.  Can you also update README as well as `bin/start_,migration.sh` if the usage of migration has changed?  Thanks.

chitang:
I think it is a problem more than just documentation.  I will work on test-database.

velvia:
:+1:

PR #71, opened 2015-05-22
Support dynamic host port mapping under Docker.

chitang:


velvia:
@chitang reviewed, looks fine now that I understand the env var system.

PR #73, opened 2015-05-27
Specify english when querying since we specify it creating the index.

znep:
If not specified, postgres uses its default as set in default_text_search_config.  In RDS this is
set to simple whereas in other environments it gets set to english based on the locale.

http://www.postgresql.org/docs/9.4/static/runtime-config-client.html#GUC-DEFAULT-TEXT-SEARCH-CONFIG

The result was, for example, 'vancouver' getting indexed as 'vancouv' but the same wasn't happening
at query time so no results were found.


sokrahta:
:shipit:

PR #76, opened 2015-06-04
Refresh rollup because data version changes.

chitang:


znep:
We should review SnapshotDropped to ensure it isn't updating the version on the copy, but this looks like it should fix the specific problem we ran into.

PR #79, opened 2015-06-16
Run PG secondary migrations on container start

urmilan:


znep:
I'm not sure if this will work because soql-server-pg has read only access to the database last I knew, and doesn't have permission to make schema changes... so either we need to change that or do it on the secondary watcher side in a loop.

chitang:
Is it always safe to run migration?  How do we handle migrations which require locks on tables?

znep:
Don't deploy the code until you can run the migration.  If that means
delaying merging it to master, then that is what it means.

The goal is to not be in a constant state of uncertainty about what
migrations have or haven't been run.

PR #81, opened 2015-07-10
Allow overriding postgres work_mem setting.

znep:
This is to allow us to bump up this setting for production environments
with sufficient memory available in order to speed up rollup table
creation by using a HashAggregate instead of a GroupAggregate.


urmilan:
:shipit: if the test failures are spurious

PR #83, opened 2015-07-17
static analysis

sokrahta:
* upgrade sbt to 0.13.8
* replace socrata-cloudbees-sbt with socrata-sbt-plugins
* address some compile errors
* replace `Version` generated code with `BuildInfo`
* addressed ~450 style warnings/errors


rozap:
this looks good to me :shipit:

PR #85, opened 2015-07-20
Change from implicit conversion to implicit parameter.  Create less o…

chitang:
…bjects.


common-pg/src/main/scala/com/socrata/pg/soql/ExprSqlizer.scala
```diff
@@ -35,19 +37,22 @@ class StringLiteralSqlizer(lit: StringLiteral[SoQLType]) extends Sqlizer[StringL
     s"e'${escape(s)}'"
   }
 
-  private def toUpper(lit: String, ctx: Context): String = if (useUpper(ctx)) lit.toUpperCase else lit
+  private def toUpper(lit: StringLiteral[SoQLType])(v: String, ctx: Context): String =
```
kaidamasaki: Consider making these a single parameter list, or making `lit` an implicit argument.

kaidamasaki:
Looks good to me.

PR #87, opened 2015-07-30
Make JMX work on dynamic port.

chitang:


znep:
How does this work?  For dynamic ports you need it to listen on one port and advertise on another, no?

chitang:
Marc, you are right.  I forgot how it works.  But right now it isn't working.  May be JMX needs some extra port?

znep:
Yeah... I don't actually know how to make it work offhand.  I had thought
there were code changes required to have a custom (something) to advertise
a different port than it listened on but that was long ago...

PR #88, opened 2015-08-05
Nathaliek/core 5400/add table cleanup delay test

NathalieKaligirwa:
@urmilan 
This branch has the most up to date changes. The name of the branch is misleading (I was experimenting with a new gui thus the 'test' part). 
Changes in soql-postgres-adapter: 

- make sure all functions involved in dataset deletion in postgres-adapter are up to date.


NathalieKaligirwa:
Jenkins is failing because the version of dc is not available yet. On release, add the correct dc dependency.

urmilan:
I have the same feedback here as for the data-coordinator pull request. We should update `SqlDatasetDropper` to delete both t-tables and metadata, and then remove all table dropper thread code in the secondary.

urmilan:
Reviewed up to f1c4587

urmilan:
I _think_ once we remove the `tableCleanupDelay` config in coordinator-lib, the only change needed in this repo will be the dependency bump?

PR #89, opened 2015-09-08
Add curated_region_test function.

chitang:


common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -276,4 +278,14 @@ object SqlFunctions {
     val s = s"$l $fnName $r"
     ParametricSql(s, setParamsLR)
   }
-}
+
+  private def curatedRegionTest = {
+    formatCall(
+      """case when st_npoints(%s) > %s then 'too complex'
+              when st_xmin(%s) < -180 or st_xmax(%s) > 180 or st_ymin(%s) < -90 or st_ymax(%s) > 90 then 'out of bound'
```
urmilan: 'out of bound**s**' (plural)

common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -276,4 +278,14 @@ object SqlFunctions {
...
+              when not st_isvalid(%s) then st_isvaliddetail(%s)::text
```
urmilan: Playing around in my local PostGIS instance, this returns a difficult-to-parse output to the user (including the offending shape in Postgis-Binary :
```
localhost:5432 blist@test=# SELECT st_isvaliddetail(the_geom)::text FROM .....
                         st_isvaliddetail
------------------------------------------------------------------
 (f,Self-intersection,010100000000000000000040400000000000001440)
 (f,Self-intersection,01010000000000000000004540FFFFFFFFFFFF1340)
 (f,Self-intersection,01010000000000000000004A40FFFFFFFFFFFF1340)
```

Can we return something a little more useful in the output? For example
```
localhost:5432 blist@test=# SELECT reason(ST_IsValidDetail(the_geom)) as reason, ST_AsGeoJson(location(ST_IsValidDetail(the_geom))) as location FROM .....
      reason       |               location
-------------------+---------------------------------------
 Self-intersection | {"type":"Point","coordinates":[32,5]}
 Self-intersection | {"type":"Point","coordinates":[42,5]}
 Self-intersection | {"type":"Point","coordinates":[52,5]}
(3 rows)
```

urmilan:
Reviewed up to 91656eb

urmilan:
:shipit: once the build issues are fixed.

PR #97, opened 2015-10-05
Add simplify_preserve_topology

kaidamasaki:


urmilan:
Pls add a test for the new function to `SqlizerTest`... apart from that :shipit:

PR #98, opened 2015-10-06
Fix all the scalastyle errors, because I'm neurotic

kaidamasaki:


soql-server-pg/src/test/scala/com/socrata/pg/server/SoQLStringFunctionsTest.scala
```diff
@@ -44,6 +44,7 @@ class SoQLStringFunctionsTest extends SoQLTest {
   }
 
   test("contains(c, x)") {
-    println("Search should be used instead of contains.")
+    // ...What even is this for?
```
znep: It is a function to find a substring in a string that has been used in the past to emulate a search.  But it is slow and horrible so we don't support it.  Well... we implement the function just in a way that doesn't actually do contains just equality.  Ideally we would just error out and the test could check that...

znep:
:ship:    Sorry for the delay

Looks like it has some small merge conflicts that need to be fixed before it can be merged.

PR #102, opened 2015-10-27
Enable blob type

rjmac:
Blobs are represented by a text column, and there are only two functions
on them: comparing them for equality, and an identity-ish conversion of
text to blob (which is implicitly used on literal string to convert them
to blob values if required).


chitang:
But:

blist_test=# select * from 
blist_test-# (
blist_test(# select 'A Z'
blist_test(# union all
blist_test(# select 'APB'
blist_test(# union all
blist_test(# select 'a b'
blist_test(# union all
blist_test(# select 'apb'
blist_test(# ) t order by 1
blist_test-# ;
 ?column? 
----------
 a b
 apb
 APB
 A Z
(4 rows)

And tests are failing on my box.  May be change space to some letter for now and investigate this later?

chitang:
In <, looks like postgres removes space:

blist_test=# select 'APCO' < 'A M', 'APCO' < 'A Q';
 ?column? | ?column? 
----------+----------
 f        | t
(1 row)

PR #108, opened 2015-12-31
Handle nulls in location sub-columns.

chitang:


common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -21,6 +21,11 @@ object SqlFunctions extends SqlFunctionsLocation {
     (FunCall, Map[UserColumnId, SqlColumnRep[SoQLType, SoQLValue]], Seq[SetParam], Sqlizer.Context, Escape)
       => ParametricSql
 
+  val SqlNull = "null"
+  val SqlNullInParan = "(null)"
```
meganschneider: nit: typo? not sure if paran is paren or param so far..

meganschneider:
:shipit: (assuming tests pass and whatnot)

PR #110, opened 2016-01-07
Make the tests be more flexible about floating-point numbers

rjmac:
By accepting a number if it's within 0.001% of the expected value


kaidamasaki:
Beautiful.

PR #113, opened 2016-01-16
EN-158: Updates to use new version of data-coordinator

aerust:
Bumps to v.2.0.0 of data-coordinator.
 - BD migrations need to be run, but should otherwise be releasable
 - Need to write tests


aerust:
- uses v2.1.3 of data-coordinator
- unit tests are added
- a migration (included) will need to be run on release

PR #116, opened 2016-01-22
EN-1783 Make money rep represent money type.

chitang:


chitang:
Oops, there are more issues to be fixed.

PR #119, opened 2016-01-28
EN-2137: Properly log cause of being unable to populate a rollup.

znep:
Typically these are parsing errors if the schema has changed and the rollup hasn't been updated, eg.
delta importer drops and re-adds columns on schema changes and they get different internal ids.

This is a partial revert of 6c433194 which introduced this bug in a code review style change.  The issue
was it resulted in e not being typed as a Throwable, so it instead was calling an overloaded
warn(message : scala.Predef.String, params : scala.AnyRef*) method which then effectively ignored the
params argument.


urmilan:
Would it be helpful for the warning message to log whether it was a `SoQLException` or a `StandaloneLexerException`? Aside from that :shipit:

PR #120, opened 2016-02-18
These are problematic when we start using chained soql.

chitang:
Only add ST_AsBinary in outermost soql.
Quote identifiers more aggressively.


AEscobarCruz:
lgtm :shipit:

PR #122, opened 2016-02-22
Update for new soql typechecking

rjmac:


chitang:
lgtm, I updated some failed tests because the soql compile error message is slightly different.

rjmac:
Ah, thanks for doing that!

PR #123, opened 2016-02-26
EN-3281 - Consider discarded copies when determining dataVersion delta between truth and PG secondary

urmilan:
See https://socrata.atlassian.net/browse/EN-3281

We are currently triggering unnecessary resyncs on the version event immediately following the dropping of a working copy. This is because we are using `PostgresDatasetMapReader.latest` to determine the list of copies that have been replicated to the secondary, and this method excludes discarded working copies whose dropCopy event has in fact been synced to the secondary. This commit makes sure that we consider those discarded copies when deciding whether to resync.


urmilan:
@znep @rjmac @chitang Would appreciate your eyes on this one.

rjmac:
I _think_ this is ok; :shipit: carefully.

PR #125, opened 2016-03-17
EN-3976 - pg-adapter implementation of SoQL num_points function

urmilan:


urmilan:
Jenkins failures are due to the snapshot dependency on soql-stdlib. Should pass once we use a release build.

chitang:
Is it missing a text file?  Otherwise, lg

urmilan:
Yes it is, good catch Chi. Thanks.

PR #126, opened 2016-03-23
Add SoQL functions point.latitude, point.longitude

chitang:
Won't build until we bump soql-reference.


meganschneider:
:shipit: (test fails just versioning?)

PR #127, opened 2016-03-24
EN-3726: function to validate geometries

NathalieKaligirwa:


AEscobarCruz:
:shipit: Keeping in mind there is some questions about soql analyzer.

PR #128, opened 2016-03-28
Add soql functions to construct location and phone from simple values.

chitang:


common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctionsLocation.scala
```diff
@@ -24,6 +24,15 @@ trait SqlFunctionsLocation {
...
     LocationToAddress -> locationAddress _,
+    // TODO: human_address will be null if any address field is null.  But we cannot do a simple coalesce
+    // TODO: without handling the edge case to keep human_address as null when all address fields are indeed null.
+    // The line break in the verbatium string is significant which keep the two sub-fields as
+    // two separate fields like (f1), (f2) instead of (f1, f2)
+    // TODO: May be we should not wrap expression in () to begin with.
+    Location -> formatCall(
+      "ST_AsBinary(%s)" +
+      SqlFragments.Separator +
+      """'{"address": "' || %s ||  '", "city": "' || %s || '", "state": "' || %s || '", "zip": "' || %s || '"}'""".stripMargin) _,
```
rjmac: If this is constructiong JSON (and it looks like it) it'd be preferable to use the JSON string interpolator and `CompactJsonWriter.toString` to string concatenation.

rjmac: Hm actually it's construction a SQL string literal containing JSON?  Ick.  Is there any way to do this with a prepared statement placeholder instead?  What happens if one of the "%s"es contains a single-quote?

common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -262,7 +262,7 @@ object SqlFunctions extends SqlFunctionsLocation with SqlFunctionsGeometry with
       case Some(op) =>
         foldSegments(sqlFragsAndParams._1.map(s => template.format(s)), op)
     }
-    ParametricSql(Seq(foldedSql), sqlFragsAndParams._2)
+    ParametricSql(foldedSql.split(SqlFragments.SeparatorRx).toSeq, sqlFragsAndParams._2)
```
rjmac: Newlines are allowed in our input... I'm not sure there's any sort of in-band separation that's safe to use.

chitang: I think what matters is that the template in formatCall do not collide with the separator.  User inputs should be using place holders and parameters.  They don't appear in the sql.

PR #131, opened 2016-04-04
added soqlfunctions from soql ref

AEscobarCruz:


chitang:
lgtm but should add some tests.

AEscobarCruz:
next release will do. just want to unblock release for now.

PR #133, opened 2016-04-05
EN-4519: PG Secondary resync attempts to drop the wrong set of rollup…

urmilan:
…s in some cases

### DIFF
--- a/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
+++ b/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
@@ -475,7 +475,7 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
     // re-create rollup metadata
     for { rollup <- rollups } RollupCreatedOrUpdatedHandler(pgu, postUpdateTruthCopyInfo, rollup)
     // re-create rollup tables
-    updateRollups(pgu, postUpdateTruthCopyInfo, dropFirst = false)
+    updateRollups(pgu, postUpdateTruthCopyInfo, dropFirst = true)
 
     cookie
   }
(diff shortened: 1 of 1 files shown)


store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -475,7 +475,7 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
     // re-create rollup metadata
     for { rollup <- rollups } RollupCreatedOrUpdatedHandler(pgu, postUpdateTruthCopyInfo, rollup)
     // re-create rollup tables
-    updateRollups(pgu, postUpdateTruthCopyInfo, dropFirst = false)
```
AEscobarCruz: Would be good to a note as to why this is important, since you did mention in slack it was time consuming to get to this conclusion.

znep:
Hrm... so now the dropFirst param to updateRollups is totally unused?  If so we should probably remove it entirely.  I can't see what this change could harm as we are already saying "drop if exists" ... but am still somewhat confused by the whole resync codepath... it feels like there are probably other gotchas lurking here.  :(

Did you check with @chitang about why it was added in the first place in a5d87156012540f7326d8a35e1b170000e5d7c54?  

That said, :ship: :boat: pending the above questions.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
urmilan: The time spent was mainly to investigate and repro the bug, which is summarized here. https://socrata.atlassian.net/browse/EN-4519

chitang:
I think resync clean up everything in prior. The parameter is there to safe some unnecessary work.  But since it is indempotent, it is ok to simplify it.

urmilan:
Removed the dropFirst parameter per discussion with @chitang and @znep. Will merge once PR builder builds green again.

PR #134, opened 2016-04-07
Implement humanAddress which was kind of already implemented.  Refact…

chitang:
…or to share implementation.


aerust:
looks good to me

PR #136, opened 2016-04-13
EN-4323: Implement Secondary.dropCopy(.)

aerust:
Secondary.dropCopy(.) is part of the resync
path for discarded and snappshotted copies.
Keep an entry in `copy_map` for a discarded
copy during resync only if it is the latest
copy of the dataset.

We want to do this because secondary-watcher
updates the secondary manifest to say that
a secondary is done with the _latest_ copy
of a dataset after resync.

### DIFF
--- a/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
+++ b/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
@@ -129,8 +129,11 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
   def dropCopy(datasetInfo: DatasetInfo,
                secondaryCopyInfo: SecondaryCopyInfo,
                cookie: Secondary.Cookie, isLatestCopy: Boolean): Secondary.Cookie = {
-    withPgu(dsInfo, None) { pgu =>
+    logger.info("droping copy (datasetInfo: {}, secondaryCopyInfo: {}, cookie: {})",
+      datasetInfo, secondaryCopyInfo, cookie)
+    withPgu(dsInfo, Some(datasetInfo)) { pgu =>
       doDropCopy(pgu, datasetInfo, secondaryCopyInfo, isLatestCopy)
+      pgu.commit()
     }
     cookie
   }
--- a/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
+++ b/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
@@ -49,14 +49,6 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
     logger.info("shut down {} {}", dsConfig.host, dsConfig.database)
   }
 
-  // Return true to get all the events from the stream of updates from the data-coordinator
-  // Returning false here means that instead of a stream of updates from the DC, we will receive
-  // the resync event instead.
-  def wantsWorkingCopies: Boolean = {
-    logger.debug("wantsWorkingCopies returning true")
-    true
-  }
-
   // This is the last event we will every receive for a given dataset. Receiving this event means
   // that all data related to that dataset can/should be destroyed
   def dropDataset(datasetInternalName: String, cookie: Secondary.Cookie): Unit = {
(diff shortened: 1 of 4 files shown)


aerust:
Still need to test this.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -133,23 +125,64 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
 
+  def doDropCopy(pgu: PGSecondaryUniverse[SoQLType, SoQLValue], datasetInfo: DatasetInfo,
+                 secondaryCopyInfo: SecondaryCopyInfo, isLatestCopy: Boolean): Unit = {
+    // If this is the latest copy we want make sure this copy is in the `copy_map`.
+    // (Because secondary-watcher will write back to the secondary manifest that we
+    // are up to this copy's data-version.)
+    // If this is the case, resync(.) will have already been called on a (un)published
+    // copy and appropriate metadata should exist for this dataset.
+    // Otherwise, we want to drop this copy entirely.
+    val truthCopyInfo = pgu.secondaryDatasetMapReader.datasetIdForInternalName(datasetInfo.internalName) match {
```
chitang: You don't need the val truthCopyInfo, do you?

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -133,23 +125,64 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
+          logger.error("No dataset found for {} when dropping the latest copy {} for data version {}?",
+            datasetInfo.internalName, secondaryCopyInfo.copyNumber.toString, secondaryCopyInfo.dataVersion.toString)
+      case Some(dsId) =>
+        pgu.datasetMapReader.datasetInfo(dsId).map { truthDatasetInfo =>
+          pgu.datasetMapWriter.copyNumber(truthDatasetInfo, secondaryCopyInfo.copyNumber) match {
+            case Some(copyInfo) =>
+              // drop the copy
+              logger.info(
+                "dropping copy {} {}",
+                truthDatasetInfo.systemId.toString(),
```
chitang: remove () after toString.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -49,14 +49,6 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
     logger.info("shut down {} {}", dsConfig.host, dsConfig.database)
   }
 
-  // Return true to get all the events from the stream of updates from the data-coordinator
```
chitang: Why don't we need this any more?

aerust: All the secondaries return `true`,  I removed it because it was a fairly complicated unused code path.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
aerust: Ah right, nope. Unless there is something more I am forgetting to do?

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
chitang: As long as Robert is OK with this.  There was a ES secondary that did not want working copy.

chitang:
May be good to move dropCopy to a handler class and set up a test to check the version?

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
aerust: Robert was, but the spandex secondary currently gets working copies, which seems weird.

aerust:
Sounds like a good idea (would be nice to have at least one unit test around this)

store-pg/src/test/scala/com/socrata/pg/store/ResyncTest.scala
```diff
@@ -69,5 +69,101 @@ class ResyncTest extends PGSecondaryTestBase with PGStoreTestBase with PGSeconda
...
 
+  test("handle drop copy") {
+    withPgu() { pgu =>
+      val pgs = new PGSecondary(config)
+      val secondaryDatasetInfo = DatasetInfo("monkey", "locale", "obfuscate".getBytes)
+
+      val snapshottedCopy = CopyInfo(new CopyId(123), 1, LifecycleStage.Snapshotted, 24, new DateTime())
+      val publishedCopy = CopyInfo(new CopyId(123), 2, LifecycleStage.Published, 55, new DateTime())
+      val unpublishedCopy = CopyInfo(new CopyId(123), 3, LifecycleStage.Unpublished, 57, new DateTime())
+      val discardedCopy = CopyInfo(new CopyId(123), 3, LifecycleStage.Discarded, 58, new DateTime())
```
chitang: Would the discarded copy number be 4 instead of 3 in real case?

chitang:
lgtm if in real case the copy number is 4.

store-pg/src/test/scala/com/socrata/pg/store/ResyncTest.scala
aerust: Yeah, when you discard a working copy it maintains its copy number. Or that was what I was trying to represent with my test.

chitang: OK, i was thinking more like repeatedly create working copies and discard them.

chitang:
lgtm

PR #141, opened 2016-04-19
When I try to bump DC to v2.1.14-SNAPSHOT, a test suddenly fails.

chitang:
It was succeeding only because a map is traversed at a specific order and return the "right" key columns from two conflicting columns.
This actually removes the duplicated key column.


rjmac:
:+1:

PR #146, opened 2016-04-25
EN-5261: Fix most-recent copy related bugs.

aerust:
Fix false assumptions around:
* most-recently-created copy is the most-recent copy

Pickup secondary-watcher changes


rjmac:
:+1:

PR #148, opened 2016-04-28
When pg secondary is in a state where newer copies exist but is asked…

chitang:
… by SW to play version of some older copy, it will fail (but won't kill SW).

This helps get out of the corrupted situation by kicking off resync.


rjmac:
ok

PR #150, opened 2016-05-03
En 2888

rjmac:
The soql-postgres-adapter half of this thing.


store-pg/src/main/scala/com/socrata/pg/store/events/RowsChangedPreviewHandler.scala
```diff
@@ -0,0 +1,43 @@
...
+      if(!truncated) {
+        val copier = pgu.datasetContentsCopier(logger)
+        val schema = pgu.datasetMapReader.schema(newTruthCopyInfo)
+        val copyCtx = new DatasetCopyContext[SoQLType](newTruthCopyInfo, schema)
+        copier.copy(truthCopyInfo, copyCtx)
+      }
+
+      schema.values.find(_.isSystemPrimaryKey).foreach(schemaLoader.makeSystemPrimaryKey)
+
+      pgu.tableDropper.scheduleForDropping(truthCopyInfo.dataTableName)
```
chitang: Do we currently interfere with query while the secondary copy is being updated?  Do we need the ability to delay table drop until the playback is complete?  There can be reindexing and rollup at the end.  They can take a long time to process.

rjmac: That's why it just schedules it for dropping; this puts the table name in a list, and an asynchronous process comes by later to actually drop it.

chitang:
lgtm if the interference between update and query does not get worse.

store-pg/src/main/scala/com/socrata/pg/store/events/RowsChangedPreviewHandler.scala
chitang: OK, I think actual clean up already has a buffer.

rjmac:
Hrm, this change has broken pg-secondary tests.  I think it's related to the switch from ordering by `copy_number` to `data_version` in some queries.

PR #152, opened 2016-05-17
EN-5912: Use Java8 base image for secondary-watcher-pg

urmilan:
Looks good so far in Jenkins and staging. Will monitor in staging for a bit before merging.


urmilan:
retest this please

PR #153, opened 2016-05-17
Drop old rollup tables when doing a table-replacing upsert

rjmac:
Otherwise they just hang around forever; new rollups will be
created for the new table.


chitang:
lgtm

PR #154, opened 2016-05-20
Move the st_asbinary tweak from general conversion to specific conver…

chitang:
…sion where it belongs.

This fixes:

curl -v -k "https://domain/id/8fe6-dei9.json?$query=select * where within_circle(location,1.1,2.2,100) limit 1&$$version=2.0&$$read_from_nbe=true"


rjmac:
ok

PR #155, opened 2016-05-23
Add logging around various parts of the log playback

rjmac:
..when a table-copy happens because of a large upsert.

The cause of the table copy and times for its creation and re-indexing
and re-rolling-up are logged.


znep:
Looks good, 🚢 once the build finishes and succeeds.

PR #157, opened 2016-05-27
EN-5371: Skip null sub column

chitang:
This is done to make it easy to work with obe
select * where phone = '4251234567' will only compare phone.number
select * where phone = 'Home: 4251234567' will compare both


rjmac:
ok

PR #158, opened 2016-06-01
Squeeze all searches to the first chained soql.

chitang:


rjmac:
@chrismetcalf -- this makes me sad, are you aware and do you approve?

PR #159, opened 2016-06-01
Chained SoQL search needs to run in the context of the previous SoQL.

chitang:
When no columns are searchable, return false instead of not having a condition.


rjmac:
Ok, that looks reasonable (barring performance, but at least it should have the right semantics...)

PR #162, opened 2016-06-23
BUG: EN-3373 date_trunc_ym(floating) = '2016-01' should work

chitang:


meganschneider:
once tests pass

meganschneider:
Doesn't this have a broken test still?

PR #163, opened 2016-06-29
En 7531 - runtime-configurable thresholds for table-copying

rjmac:
Since we don't actually know what our thresholds should be for making an indexless copy on large upserts, this change allows us to store Clojure functions in Zookeeper and update them at runtime to configure the behavior based on number of columns.


chitang:
lgtm

PR #165, opened 2016-06-30
Prepare to configure Zookeeper in Marathon

rjmac:


store-pg/docker/secondary.conf.j2
```diff
@@ -46,4 +46,15 @@ pg-secondary-config {
...
+  {% if ZOOKEEPER_ENSEMBLE is defined -%}
+  curator = {
+    ensemble = {{ ZOOKEEPER_ENSEMBLE }}
+    session-timeout = 40s
+    connect-timeout = 5s
+    base-retry-wait = 100ms
+    max-retry-wait = 10s
+    max-retries = 29
+  }
+  {%- endif %)
```
chitang: Should it be {% and %}?

store-pg/docker/secondary.conf.j2
```diff
@@ -46,4 +46,15 @@ pg-secondary-config {
   }
 
   tablespace = "{{ PG_SECONDARY_TABLESPACE_FN }}"
+
+  {% if ZOOKEEPER_ENSEMBLE is defined -%}
```
chitang: Should there be a - ?

rjmac:
In both those places, the `-` is correct.  See for example [the query-coordinator config file](https://github.com/socrata-platform/query-coordinator/blob/master/query-coordinator/docker/query-coordinator.conf.j2).

aerust:
Seems reasonable to me

PR #166, opened 2016-06-30
The previous way works in query-coordinator; why didn't it work here?

rjmac:


chitang:
This makes more sense any way.  {% -%} {%- %} is very weird.

PR #176, opened 2016-08-08
EN-8957: Add a DB connection for the table dropper

aerust:


znep:
Maybe add a comment in the file about what the "+ 1" is for?  Other than that 🚢

PR #178, opened 2016-08-10
move log statement to be logged with error cases; is covered on succe…

meganschneider:
…ss case by other log messages


meganschneider:
franklin says :shipit:

PR #181, opened 2016-08-11
EN-1626: set local statement_timeout based on queryTimeoutMs param, i…

meganschneider:
…f given

### DIFF
--- a/soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
+++ b/soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
@@ -9,6 +9,7 @@ import com.socrata.datacoordinator.util.CloseableIterator
 import com.socrata.pg.soql.ParametricSql
 import com.socrata.soql.collection.OrderedMap
 import com.socrata.soql.SoQLAnalysis
+import scala.concurrent.duration.Duration
 import com.typesafe.scalalogging.slf4j.Logging
 import java.sql.{Connection, ResultSet, SQLException}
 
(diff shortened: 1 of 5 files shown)


soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
```diff
@@ -69,8 +71,23 @@ trait DataSqlizerQuerier[CT, CV] extends AbstractRepBasedDataSqlizer[CT, CV] wit
...
+      stmt.execute(s)
+    }
+  }
+
+  def setTimeout(timeoutMs: String) = s"SET LOCAL statement_timeout TO $timeoutMs"
+  def resetTimeout = "SET LOCAL statement_timeout TO DEFAULT"
+
+  private def executeSql(conn: Connection, pSql: ParametricSql, timeout: Option[Duration]): ResultSet = {
     try {
+      if (timeout.isDefined && timeout.get.isFinite()) {
```
aerust: Why are you checking that is is finite here? (just curious)

aerust:
lgtm :fish:

soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
meganschneider: is not infinite or NaN :) (also cause that seems to be our common check for "hey is this a real thing", sprinkled through code, but mostly the first bit)

meganschneider: Although that may be covered by the min/max thing below. Unsure.

aerust: Okay cool, was curious if there was some option hidden to have no timeout set.

PR #182, opened 2016-08-11
Warn about dropping nonexistant copies, rather than bailing

rjmac:
This is probably fallout from the desecondarification of snapshots.

### DIFF
--- a/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
+++ b/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
@@ -480,11 +480,15 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
         //   DELETE from falth.dataset_internal_name_map
         //    WHERE dataset_internal_name = ? -- 'alpha.20'
         logger.info("re-creating secondary dataset with new id")
-        val newCopyInfo = pgu.datasetMapWriter.create(secondaryDatasetInfo.localeName)
-        val newDatasetId = newCopyInfo.datasetInfo.systemId
+        val newDatasetInfo = pgu.datasetMapWriter.unsafeCreateDatasetAllocatingSystemId(secondaryDatasetInfo.localeName, secondaryDatasetInfo.obfuscationKey)
+        val newDatasetId = newDatasetInfo.systemId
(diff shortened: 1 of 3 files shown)


chitang:
lgtm

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -487,11 +487,15 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
         //   DELETE from falth.dataset_internal_name_map
         //    WHERE dataset_internal_name = ? -- 'alpha.20'
         logger.info("re-creating secondary dataset with new id")
-        val newCopyInfo = pgu.datasetMapWriter.create(secondaryDatasetInfo.localeName)
-        val newDatasetId = newCopyInfo.datasetInfo.systemId
+        val newDatasetInfo = pgu.datasetMapWriter.unsafeCreateDatasetAllocatingSystemId(secondaryDatasetInfo.localeName, secondaryDatasetInfo.obfuscationKey)
```
chitang: Hm... there is no unsafeCreateDatasetAllocatingSystemId

rjmac: Not yet!  There is in the companion change to data-coordinator.

chitang: ok, lgtm

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -364,7 +364,14 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
             case Some(ci) =>
               CopyDroppedHandler(pgu, ci)
             case None =>
-              throw new Exception("cannot find copy to drop")
+              // ok, we can't find the exception to drop.  This is probably down
```
znep: Need to figure out if we update this comment to reflect this actually being a temporary workaround for the other bug or figure out another approach to making the data right.

znep:
Looks good, pending bumping dc and tests passing and my one comment.

rjmac:
I've bumped DC and removed the change that would make the replication ignore nonexistant snapshots; merging as soon as jenkins is happy

PR #187, opened 2016-09-07
System PK index wasn't getting created in all circumstances

rjmac:
..leading to slow updates on large datasets

This isn't in `optimize` because that pretty much assumes it's creating
userspace indexes.

### DIFF
--- a/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
+++ b/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
@@ -551,6 +551,9 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
         loader.flush()
       }
     }
+    truthSchema.values.find(_.isSystemPrimaryKey).foreach { pkCol =>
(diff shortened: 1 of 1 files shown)


rjmac:
retest this please

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -551,6 +551,9 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
         loader.flush()
       }
     }
+    truthSchema.values.find(_.isSystemPrimaryKey).foreach { pkCol =>
+      sLoader.makeSystemPrimaryKey(pkCol)
+    }
```
chitang: Anything needs to be done if the dataset has custom user key?

rjmac: No, AFAICT the soql-pg-adaptor doesn't do anything special for user keys at all.

chitang:
lgtm

PR #191, opened 2016-10-13
EN-11200 digest soql before using it in etag.

chitang:


soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
```diff
@@ -141,7 +142,12 @@ class QueryServer(val dsInfo: DSInfo, val caseSensitivity: CaseSensitivity) exte
     // Upstream components may pass through etag headers to and from query server but generate different queries.
     // For example, "select *" may become "select `a`" or "select `a`, `newly_unhidden_colujmn`"
     // Including query string using etagInfo in hash generation makes etags more robust.
-    val etagContents = s"${datasetInternalName}_${copy.copyNumber}_${copy.dataVersion}_${etagInfo.getOrElse("")}"
+    val etagInfoDigest = etagInfo.map { x =>
+      val md = MessageDigest.getInstance("SHA1")
+      md.update(x.getBytes(StandardCharsets.UTF_8))
+      md.digest().toString
```
rjmac: digest returns a byte array; its toString method isn't useful.

PR #196, opened 2016-11-08
EN-11848: Add back info level rollup timing logging.

znep:
This got accidentally removed when we bumped the less useful timing
reports down to debug level.


aerust:
:ship:

PR #197, opened 2016-11-16
EN-4308: Use improved secondary-watcher version

aerust:
Secondary-watcher now has:
 - retry logic around releasing claims
 - health check on updating claimed_at
 - in-memory working set to only update
   claimed_at time for datasets that
   the secondary is actually still working on


chitang:
:shipit:

PR #198, opened 2016-12-08
EN-12544: Actually set work_mem for soql-server-pg.

znep:
Setting onAcquire doesn't actually do anything unless you have a connection customizer configured.


chitang:
:shipit:

PR #200, opened 2016-12-31
Chi/en 12742/primary key

chitang:
Ensure primary key


chitang:
We don't need this any more.

PR #202, opened 2017-01-24
EN-13224 Disable SoQL reads during resync.

chitang:
Resync imposes lock conflict with reads which leads to backing up of connections.  That eventually takes down a pg query server.

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
@@ -4,11 +4,12 @@ import com.rojoma.simplearm.util._
 import com.socrata.datacoordinator.id.{CopyId, DatasetId}
 import com.socrata.datacoordinator.id.sql._
 import com.socrata.datacoordinator.truth.DatabaseInReadOnlyMode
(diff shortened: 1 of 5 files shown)


common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
```diff
@@ -25,6 +26,7 @@ class PGSecondaryDatasetMapWriter[CT](override val conn: Connection,
       |VALUES (?, ?)
     """.stripMargin
   private val deleteQueryInternalNameMap = "DELETE FROM dataset_internal_name_map where dataset_system_id = ?"
+  private val enableQueryInternalNameMap = "UPDATE dataset_internal_name_map SET disabled = ? WHERE dataset_system_id = ?"
```
aerust: What was disabled previously used for?

chitang: Disabled was newly introduced last sprint.  It allows us to reject soql reads.  Write should be able to continue unaffected.

aerust: Okay, cool

aerust:
Well, lgtm :shell:

PR #204, opened 2017-01-26
EN-13224 bump dc version

chitang:


aerust:
Commit should probably also reference EN-13225 and add bullet point of
3. secondary-watcher updates version on the secondary manifest in a new transaction

store-pg/src/test/scala/com/socrata/pg/store/DatabaseTestBase.scala
```diff
@@ -158,7 +158,7 @@ trait DatabaseTestBase extends Logging {
       if (isNew) secondaryMfst.addDataset(storeId, datasetId)
       secondaryMfst.claimDatasetNeedingReplication(storeId, claimantId, claimTimeout).foreach { job =>
         try {
-          pb(NamedSecondary(storeId, secondary), job)
+          pb(NamedSecondary(storeId, secondary, ""), job)
         } finally {
```
aerust: nit: could put "read" instead of "" since it is what we use in prod for pg

chitang: really, it isn't pg in prod?

aerust:
other than nit, looks good to me :shell:

PR #206, opened 2017-02-21
EN-12729: Avoid deadlocks with new SW lib version

aerust:
Avoid deadlocks when updating multiple rows on the
`secondary_manifest` by first obtaining locks on
the rows in a deterministic order, i.e. ordery by
`dataset_system_id` then `store_id`.


chitang:
🚢

PR #208, opened 2017-03-02
EN-13048 Use group by position if we find literal

chitang:
This fixes soql like
  SELECT ‘literal’ as x GROUP BY x


aerust:
lgtm :sheep:

PR #209, opened 2017-03-14
EN-13048 Cannot use parameter for column position in order by.

chitang:
Otherwise, it will be interpreted as number literal, not column position.


znep:
Is this something we can add a unit test for?

chitang:
Actually, I don't plan to merge this.  I don't want to support order by column position yet.

PR #211, opened 2017-03-29
EN-14942 A text cell longer than 1MB causes error in creating the FTS…

chitang:
… index and stops secondary replication.

Catch the error and skip the full text search index.


aerust:
I kind of think we maybe need a way to communicate to end users that we skipped some indexes and that will impact certain queries on the dataset... but this change seems necessary to me.

aerust:
Total nit pick on commit message first line being longer that 60 characters... other than that looks good to me :sheep:

chitang:
In the spirit of echoing the bug theme, too long/big.  I am going to ignore it this time :)

PR #212, opened 2017-03-29
EN-9028 Don’t use postgres 9.6 upsert yet.

chitang:


bin/run_migrations.sh
```diff
@@ -4,7 +4,7 @@ set -e
 REALPATH=$(python -c "import os; print(os.path.realpath('$0'))")
 BINDIR=$(dirname "$REALPATH")
 
-psql datacoordinator -c "insert into secondary_stores_config (store_id, next_run_time, interval_in_seconds) values ('pg', now(), 5) on conflict do nothing"
+psql datacoordinator -c "insert into secondary_stores_config (store_id, next_run_time, interval_in_seconds) SELECT * FROM (SELECT 'pg'::text as store_id, now(), 5) t WHERE store_id NOT IN (SELECT store_id FROM secondary_stores_config)"
```
aerust: Why don't we also set the group_name to 'read' while we are at it here.

chitang: i don't see "read" better than "pg".  May be pg_read?  I don't see anything wrong with pg alone.

aerust:
as longs as builds pass, lgtm :ship:

PR #217, opened 2017-04-12
EN-15144: Configurable com.socrata log level

aerust:
Make it possible to optionally configure the
log level for com.socrata components to a higher
log level in docker for dugging Eurybates messages.


chitang:
:shipit:

PR #218, opened 2017-04-12
Bump DC version so that it usees camel case in messages instead of un…

chitang:
…derscore.

Historically, we use camel case for Eurybates messages.


aerust:
lgtm :sheep:

PR #219, opened 2017-05-05
EN-14749: Fix "data too long to index" errors for row updates.

znep:
In commit 9b4871dbf8310bb5f9ccb7983b3a958afd6f729d, we previously added the functionality
to force a resync if a value is too large to index.  Going through resync, the value will be added,
the index creation will fail, and we will move on without an index on the column.

This wasn't working for updates because they are done as batch statements so you end up with a
BatchUpdateException.

Also, add some tests around these cases.  As part of that change the row update tests to use a published
dataset because unpublished datasets have fewer things (ie. indexes) on them.


chitang:
🚢

znep:
retest this please

znep:
retest this please

again.

PR #220, opened 2017-05-05
EN-16084 Minimal change to bump soql-ref version.

chitang:
We want to do this now so that we will not be blocked if there is an urgent need to bump soil-ref for non join related work.

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
@@ -19,7 +19,7 @@ import com.socrata.pg.store.index.SoQLIndexableRep
 import com.socrata.soql.{SoQLAnalysis, SoQLAnalyzer}
 import com.socrata.soql.analyzer.SoQLAnalyzerHelper
 import com.socrata.soql.collection.OrderedMap
-import com.socrata.soql.environment.{ColumnName, DatasetContext}
+import com.socrata.soql.environment.{ColumnName, DatasetContext, TableName}
 import com.socrata.soql.exceptions.{NoSuchColumn, SoQLException}
 import com.socrata.soql.functions.{SoQLFunctionInfo, SoQLTypeInfo}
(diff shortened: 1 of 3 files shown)


chitang:
The test failure is a preexisting random deadlock failure.  I restarted it and it succeeded.

common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
```diff
@@ -61,7 +62,7 @@ class RollupManager(pgu: PGSecondaryUniverse[SoQLType, SoQLValue], copyInfo: Cop
    * Once we have the analyzed rollup query with column names, we need to remove the leading "_" on non-system
    * columns to make the names match up with the underlying dataset schema.
    */
-  private def columnNameRemovePrefixMap(cn: ColumnName): ColumnName = {
+  private def columnNameRemovePrefixMap(cn: ColumnName, todo: Qualifier): ColumnName = {
     cn.name(0) match {
```
aerust: can you pick a different variable name than `todo`, or at least add a comment of why it is called `"todo"`

common-pg/src/main/scala/com/socrata/soql/analyzer/SoQLAnalyzerHelper.scala
```diff
@@ -74,12 +76,17 @@ object SoQLAnalyzerHelper {
...
 
-      val a: SoQLAnalysis[UserColumnId, SoQLType] = analysis.mapColumnIds(newMapping)
+      val a: SoQLAnalysis[UserColumnId, SoQLType] = analysis.mapColumnIds(columnMapIgnoringQualifier(newMapping))
       (mappingWithNewColumns, convertedAnalyses :+ a)
     }
     analysesInColIds
   }
 
+  private def columnMapIgnoringQualifier(map: ColumnName => UserColumnId)(columnName: ColumnName, qualifier: Qualifier)
+    : UserColumnId = {
```
aerust: I think this function might read better just as `mapIgnoringQualifier`

aerust:
lgtm :ship:

PR #221, opened 2017-05-09
EN-15769: simplify magnitude calculation 

bahrieinn:
We had a pretty roundabout way of calculating magnitudes for a given row. 
This refactor aims to get consistent magnitude buckets so that we have inclusive starts, and exclusive end ranges across the board.

```
[a,b), [b,c), [c,d) etc
```

The motivation for changing this was that we were noticing bugs in our new visualization histograms, where values were not being placed in the correct bucket. This PR tries to solve one of the several issues we have with histograms. 

It appears that the only other places on the frontend that use `signed_magnitude_linear` (besides frontend-visualization) are:
* `CardDataService.js`
* `page_metadata_manager.rb`


bahrieinn:
retest this please

bahrieinn:
Noticed test was breaking, working on it now

chitang:
test("signed_magnitude_linear(x,width)" fails

bahrieinn:
@chitang fixed related tests, mind taking another look?

chitang:
lgtm

znep:
For testing, be aware that you might get responses based on the old version of the function due to either (1) ATS caching or (2) the query being rewritten to hit a rollup.  Updating the dataset will fix up both of those.

bahrieinn:
Thanks for the warning @znep, will definitely keep that in mind

PR #224, opened 2017-05-17
EN-15769: restore magnitude calculation change after revert

bahrieinn:
I merged this too early the first time around, as there was frontend work that depends on the updated SoQL. That work is now complete and ready to be merged, but still need to coordinate with changes here: https://github.com/socrata/frontend-visualizations/pull/582


znep:
What is the deployment ordering/strategy here given we can expect to sometimes get back the old values until the dataset being queried is updated due to rollups and caching?  Have we checked query logs to make sure no one else is using this function who may be impacted?

Presumably there are also data lens changes that have to be coordinated as well?

It is kind of annoying but I wonder if this should be a new function instead if you really do need a function with different semantics than the existing one...

bahrieinn:
I actually like the idea of a new function. It would be much more straightforward without the risk to existing consumers (though I believe data lens is the only one at this point). I'll do some refactoring here

bahrieinn:
Talked to @cmlaidlaw, we are going to do a brief spike on whether we can build not ridiculous / incorrect histograms on the front-end using row-level data, and go from there. So I will pause on the idea of making a new SoQL function here.

common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -71,8 +71,8 @@ object SqlFunctions extends SqlFunctionsLocation with SqlFunctionsGeometry with
       paramPosition = Some(Seq(0,0))),
     SignedMagnitudeLinear ->
       formatCall(
-        "case when %s = 1 then floor(%s) else sign(%s) * floor(abs(%s)/%s + 1) end",
-        paramPosition = Some(Seq(1,0,0,0,1))),
+        "case when %s = 1 then floor(%s) else floor(%s/%s) end",
```
bahrieinn: Actually the `%s = 1` case is spurious. This entire statement can simply be `floor(%s/%s)`.

There was one hypothesis that the original code was protecting against division by 0 by doing a `+ 1` but that's not the case, bc it would've had to be `floor(abs(%s) / (%s + 1))`

PR #225, opened 2017-05-17
EN-15963 Fix a regression where rollup is not used when it should be.

chitang:


aerust:
Can you include in your commit message what the regression from (i.e joins work or the actual sha of the commit)

aerust:
okay :sheep:

PR #228, opened 2017-05-23
EN-16378 Handle grouping derived views.

chitang:


common-pg/src/main/scala/com/socrata/pg/soql/ExprSqlizer.scala
```diff
@@ -125,7 +125,7 @@ object ColumnRefSqlizer extends Sqlizer[ColumnRef[UserColumnId, SoQLType]] {
           escape: Escape): ParametricSql = {
     val tableMap = ctx(TableAliasMap).asInstanceOf[Map[String, String]]
     reps.get(QualifiedUserColumnId(expr.qualifier, expr.column)) match {
-      case Some(rep) if ctx(InnermostSoql) == true => // scalastyle:off simplify.boolean.expression
+      case Some(rep) if ctx(InnermostSoql) == true || expr.qualifier.nonEmpty => // scalastyle:off simplify.boolean.expression
```
aerust: Is there a reason `ctx(InnermostSoql) == true` is being used here?

aerust:
Can you explain in comment or commit message why adding `|| expr.qualifier.nonEmpty` is the right thing to do? (both for my sake and for future sake)

common-pg/src/main/scala/com/socrata/pg/soql/ExprSqlizer.scala
chitang: not inner most is handled in the next case statement.  Until joins, innermost is the only one that directly interacts with t tables.

chitang: i am adding tests related to this.

aerust: oh. more of a question of why `== true`

aerust:
Okay, once it builds, looks good to me. :sheep:

PR #229, opened 2017-05-24
EN-16378 Improve test foundation to support soql joins and chaining.

chitang:
With the improvement, we can add a group and join test.


common-pg/src/main/scala/com/socrata/soql/analyzer/QualifiedColumnName.scala
```diff
@@ -0,0 +1,5 @@
+package com.socrata.soql.analyzer
+
+import com.socrata.soql.environment.ColumnName
+
+case class QualifiedColumnName(val qualifier: Option[String], val columnName: ColumnName)
```
aerust: should have a new line afterwards

soql-server-pg/src/test/resources/fixtures/group-join.json
```diff
@@ -0,0 +1,2 @@
+[ { "make" : "OZONE", "ct" : 2, "timezone" : 2 }
+]
```
aerust: new line

PR #230, opened 2017-06-07
EN-16779: Enable parallel queries by setting fetchSize to 0

znep:
As per https://www.postgresql.org/docs/9.6/static/when-can-parallel-query-be-used.html parallel
query can not be used if:

  The client sends an Execute message with a non-zero fetch count.
  See the discussion of the extended query protocol. Since libpq
  currently provides no way to send such a message, this can only
  occur when using a client that does not rely on libpq. If this is
  a frequent occurrence, it may be a good idea to set
  max_parallel_workers_per_gather in sessions where it is likely, so
  as to avoid generating query plans that may be suboptimal when run
  serially.

We set a fetch size so we can stream resuls without the jdbc driver
buffering them all.  This results in the driver using a cursor which
prevents parallel query action.

However, given that most of the queries that really benefit from
parallel queries are aggregation queries, and we have a default
limit of 1000 rows returned (which just happens to be the same as
the fetch size we were using), if we know the query has a limit <=
our desired fetch size then we can just set it to 0 without worrying
about running out of memory, and let postgres use parallel queries.

### DIFF
--- a/soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
+++ b/soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
@@ -29,7 +29,8 @@ trait DataSqlizerQuerier[CT, CV] extends AbstractRepBasedDataSqlizer[CT, CV] wit
     val rowCount: Option[Long] =
       if (reqRowCount) {
         val rowCountSql = toRowCountSql(analyses, dataTableName)
-        using(executeSql(conn, rowCountSql, queryTimeout)) { rs =>
+        // We know this will only return one row, so fetchsize of 0 is ok
+        using(executeSql(conn, rowCountSql, queryTimeout, 0)) { rs =>
           try {
             rs.next()
             Some(rs.getLong(1))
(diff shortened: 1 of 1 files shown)


soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
```diff
@@ -49,9 +50,21 @@ trait DataSqlizerQuerier[CT, CV] extends AbstractRepBasedDataSqlizer[CT, CV] wit
...
 
+    /*
+     * We need to set a fetchSize on the statement to limit the amount of data we are buffering in memory at
+     * once.  Unfortunately, this currently prevents parallel query execution in postgres because it uses cursors.
+     * Since the major benefit of parallel queries is for queries that take a lot of rows and reduce them to a small number,
+     * if we can tell from the query that it can't return more than sqlFetchSize rows, we can safely set the fetchsize
+     * to 0 and get the benefits of parallel queries.
+     */
+    val fetchSize = analyses.last.limit match {
+      case Some(n) if (n <= sqlFetchSize) => 0
```
aerust: Let's say we want to turn off parallel query execution for some kind of emergency, and we do that. It will still be totally fine that we are using fetch size of 0 for limited queries (`n <= sqlFetchSize`)? Or would it be nice to be able to return to our old behavior with a marathon configuration change?

znep: We would just reconfigure the database.  Wouldn't want to rely on this quirk to disable it.

aerust: Not what I was asking, if we reconfigure the database to not have parallel query execution, this code change is fine?

znep: Yup, it just allows them to be possible in some cases.

aerust:
lgtm

PR #231, opened 2017-06-07
EN-15953/SI-5069 Add median function

chitang:


aerust:
```
[info] *** 1 TEST FAILED ***
[error] Failed tests:
[error] 	com.socrata.pg.server.SoQLMathFunctionsTest
[error] (soql-server-pg/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 28 s, completed Jun 7, 2017 7:45:59 PM
```

aerust:
We were thinking that it would make sense to include adding percentile when we add median, want to do that too?

Also where all does the soql-reference version need bumping, core and query-coordinator in addition to here?

chitang:
I was just wondering if we'd like to offer percentile only. We can deal with percentile and its variants when it is requested in the future.

Version bump in other projects will follow after this.

I need to find out why test fails though.  It runs fine locally.

znep:
We should probably update the version of postgres we run tests against if we are going to rely on newer features.

soql-server-pg/src/test/resources/fixtures/median.json
```diff
@@ -0,0 +1 @@
+[ { "m1": 50, "m2": "HOTEL"} ]
```
aerust: nit: new line after this

aerust:
Yeah, this seems like a thing that our units tests should work for and cover

PR #236, opened 2017-07-03
EN-16370 Support sub-queries in joins for derived views.

chitang:


rjmac:
Hm, that looks OK from a safety perspective.  How will that error get surfaced to the user?  (I'd really prefer a transform-to-cleaned approach, but that'll be a lot more annoying since you'd have to walk the tree to fix up all references to the cleaned aliases, and handle post-cleaning collisions.)

chitang:
Soql error is already broken in general because we manipulate user soql too much especially for derived views.
Yeah, I would like to use internal alias too.  But that can be future improvement work.  I am still leaning towards reserving sub-query joins for internal derived view only.  I probably throw an error in core if that is used directly by users.

PR #248, opened 2017-11-01
Bump DC version

chitang:


common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryRowReader.scala
```diff
@@ -16,7 +16,7 @@ class PGSecondaryRowReader[CT, CV](val connection:Connection,
 
   def lookupRows(ids: Iterator[CV]): RowUserIdMap[CV, InspectedRow[CV]] =
     timingReport("lookup-rows") {
-      using(sqlizer.findRows(connection, ids)) { it =>
+      using(sqlizer.findRows(connection, true, ids)) { it =>
         val result = datasetContext.makeIdMap[InspectedRow[CV]]()
```
aerust: `bySystemIdForced = false`

PR #251, opened 2017-11-14
EN-19155: add ST_CollectionExtract and bump version

bahrieinn:


chitang:
If postgres can take different geometry type with the same function, it is probably better if we do parameter type binding.

bahrieinn:
restest this please

PR #253, opened 2018-01-05
EN-20583: enable solo for soql-server-pg and secondary-watcher

michaelb990:


bin/build_soql_server_pg_docker.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -e
+
+#sbt soql-server-pg/assembly
```
tkbrigham: Remove comment

configs/soql-postgres-adapter.conf
```diff
@@ -0,0 +1,142 @@
+# Config for secondary-watcher and soql-server-pg
```
tkbrigham: Is there any argument for splitting these two configs?

michaelb990:
retest this please

PR #261, opened 2018-02-06
EN-17021: PG SW can help maintain dataset sizes

aerust:
secondary-watcher-pg can maintain the secondary_metrics tables for
its stores. This is for collocation so we can move smaller datasets
rather than larger datasets.

Also, include configuration so we can turn this off incase it causes
problems in production.

Also .gitignore .build.sh.pid created by the bin/start scripts.


aerust:
retest this please

PR #265, opened 2018-03-08
SPARTA: Starts scripts for solo for phonetic SWs

aerust:
Start scripts for solo for running secondary-watcher-pgs
for the phonetic truth cluster "alpha", "bravo", "charlie"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@
 */docker/*.jar
 configs/local*.conf
 .build.sh.pid
+secondary-watcher-pg*.log
+soql-server-pg*.log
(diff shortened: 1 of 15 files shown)


znep:
What are the 6300, 6500, 6700 ports and where are they configured?

aerust:
@znep 6300 is theoretically the jmx port for secondary-watcher-pg, if it works? but mostly in this context it is just a port that is open to make solo happy (and they need to be different so solo can tell the different containers apart)

znep:
If nothing is listening on the port and solo doesn't care then ... what is
the point of solo even checking it...

aerust:
@znep No the point is, solo does care. That is how solo currently can differentiate the difference between which docker container running the `secondary-watcher-pg` image is the `alpha` vrs `bravo`, vrs `charlie`

znep:
bitmoji

So it only cares that docker is configured to support a port and not that the application actually is running or listening on it?  That is ... bound to be the source of future confusion, but ok.

PR #277, opened 2018-06-01
EN-25454 Add numeric type column match to SoQL search

chitang:


common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
```diff
@@ -173,15 +173,44 @@ object SoQLAnalysisSqlizer extends Sqlizer[AnalysisTarget] {
...
+          Some(NumberLiteral(number, SoQLNumber.t)(searchLit.position))
+        } catch {
+          case _: NumberFormatException =>
+            None
+        }
+
+      val numericCols = searchNumberLitOpt match {
+        case None => Seq.empty[String]
+        case Some(searchNumberLit) =>
+          prevAna match {
```
PiSoup: What is the purpose of this match block? I see a nearly identical version of this on line 169 as well, could it be DRYed?

common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
```diff
@@ -173,15 +173,44 @@ object SoQLAnalysisSqlizer extends Sqlizer[AnalysisTarget] {
...
-      searchVector match {
-        case Some(sv) =>
-          val fts = s"$andOrWhere $sv @@ plainto_tsquery('english', $searchSql)"
-          ParametricSql(Seq(fts), searchSetParams)
-        case None =>
-          ParametricSql(Seq(s"$andOrWhere false"), setParamsWhere)
+      if (searchPlusNumericParametricSql.sql.isEmpty) {
+        ParametricSql(Seq(s"$andOrWhere false"), setParamsWhere)
+      } else {
+        ParametricSql(searchPlusNumericParametricSql.sql.map { x => s"$andOrWhere ($x)" }, searchPlusNumericParametricSql.setParams)
```
PiSoup: If `where` is defined will this end up generating SQL like `select * WHERE clause1 WHERE clause2 ...`?

common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
```diff
@@ -173,15 +173,44 @@ object SoQLAnalysisSqlizer extends Sqlizer[AnalysisTarget] {
           PostgresUniverseCommon.searchVector(primaryTableReps, Some(ctx))
       }
 
+      val searchNumberLitOpt =
+        try {
+          val number = BigDecimal.apply(searchLit.value)
+          Some(NumberLiteral(number, SoQLNumber.t)(searchLit.position))
```
PiSoup: Is `searchLit.position` always going to be `NoPosition` or is there something in the apply of `StringLiteral` on line 165?

common-pg/src/main/scala/com/socrata/pg/store/index/FullTextSearch.scala
```diff
@@ -33,6 +34,22 @@ trait FullTextSearch[CT] {
     toTsVector(cols.toSeq)
   }
 
+  def searchNumericVector(selection: OrderedMap[ColumnName, CoreExpr[_, CT]], ctx: Option[Context]): Seq[String] = {
+    val cols = selection.view.filter(x => SearchableNumericTypes.contains(x._2.typ)).map { case (columnName, expr) =>
+      val name = ctx.map(c => qualify(columnName.name, c)).getOrElse(columnName.name)
```
PiSoup: Don't really need to assign a val here. Could also use the shorthand you're using on line 48 for mapping the context for consistency.

common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
chitang: Your comment straddles two match.  The outer one chooses between search that is numeric and non numeric.  The inner one chooses between chained and not chained which has a similar block but dealing with the other part of search.

common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
chitang: $x should be something like '(to_vector(.....) [ or numeric_col1 = ? or numeric_col2 = ? ...)'
$andOrWhere should be WHERE if there is no filter and filter + search becomes WHERE $x.
$andOrWhere should be AND if there is filter and filter + search becomes WHERE filter AND $x.

common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
chitang: I don't really care between that and NoPosition.

common-pg/src/main/scala/com/socrata/pg/store/index/FullTextSearch.scala
chitang: Removed

PR #278, opened 2018-06-05
EN-25528: update solo configs

tkbrigham:


bin/start_secondary_watcher_pg_docker.sh
```diff
@@ -11,5 +11,4 @@ local_config_dir="$(dirname "$(realpath "$0")")/../configs"
 docker run \
   -e SERVER_CONFIG="configs/application.conf" \
   -v "$local_config_dir":/srv/secondary-watcher/configs \
-  -p 6300:6300 \
```
aerust: the secondary-watcher-pg doesn't have an http endpoint so solo won't know that it is running if you delete this I think? Unless solo is more advanced than I realized?

PR #288, opened 2018-10-31
Add ST_AsBinary to geometry type in group by to work around a slownes…

chitang:
…s problem


znep:
I'll need to find a bit of time to study it before I can give it a :shipit: but we should call out explicitly in the commit message and/or comment that postgis 2.4.0 appears to fix it and link to the commit https://github.com/postgis/postgis/commit/8606a3164111e75754ce59c095a05e193cdae636

znep:
retest this please

PR #289, opened 2018-11-29
EN-28710 converting geometry to text in soql query causes secondary-w…

chitang:
…atcher to fail on datasets with geometry rollup


common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
```diff
@@ -305,19 +306,23 @@ object SoQLAnalysisSqlizer extends Sqlizer[AnalysisTarget] {
         val (_, selectSetParams) = acc
         val ParametricSql(sqls, newSetParams) = Sqlizer.sql(coreExpr)(rep, typeRep, selectSetParams, ctxSelect, escape)
         val sqlGeomConverted =
-          if (ctx.contains(LeaveGeomAsIs) || strictInnermostSoql || ctx.contains(IsSubQuery)) { sqls }
-          else {
+          if (shouldConvertGeomToText(ctx) && !sqls.isEmpty) {
             val cn = if (strictOutermostSoql) Some(columnName) else None
             // compound type with a geometry and something else like "Location" type
             // must place the geometry in the first part.
-            if (false == ctx(OutermostSoql) || sqls.isEmpty) { sqls }
```
PiSoup: The movement of `false == ctx(OutermostSoql)` into the outer if statement will result in it behaving opposite to the way it previously functioned. Is that intentional?

PR #292, opened 2019-01-02
updating the reference to standard libraries to use the versions that…

JoeNunnelley:
… reference an exact soql-brita version

soql-reference@2.11.7 was changed to reference soql-brita 1.4.1 (which is in artifactory) and to remove problematic range syntax that artifactory doesn't like.  This change update the soql-postgres-adapter project to use that newly built standard lib set.


JoeNunnelley:
btw sbt test works on this branch now

PR #301, opened 2019-05-13
EN-32347 Get rid of the word "NewLoggingHandler" in logging

chitang:


soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
```diff
@@ -33,7 +33,7 @@ import com.socrata.http.server.routing.SimpleResource
 import com.socrata.http.server.routing.SimpleRouteContext._
 import com.socrata.http.server.util.Precondition._
 import com.socrata.http.server.util.RequestId.ReqIdHeader
-import com.socrata.http.server.util.handlers.{NewLoggingHandler, ThreadRenamingHandler}
+import com.socrata.http.server.util.handlers.{LoggingOptions, NewLoggingHandler, ThreadRenamingHandler}
```
PiSoup: Can the import be removed as well?

chitang: NewLoggingHandler is still used.  We just configure it not to print its name.

ThreadRenamingHandler(NewLoggingHandler(logOptions)

PR #303, opened 2019-05-24
Bump references version

chitang:


common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryUniverse.scala
```diff
@@ -91,7 +91,7 @@ class PGSecondaryUniverse[SoQLType, SoQLValue](
 
   lazy val datasetMapWriter: PGSecondaryDatasetMapWriter[SoQLType]  =
     new PGSecondaryDatasetMapWriter(conn, typeContext.typeNamespace,
-      timingReport, obfuscationKeyGenerator, initialCounterValue)
+      timingReport, obfuscationKeyGenerator, initialCounterValue, 0L)
```
rjmac: This 0 should probably come from the same place initialCounterValue comes form

common-pg/src/main/scala/com/socrata/pg/store/SoQLCommon.scala
```diff
@@ -58,6 +58,8 @@ class PostgresUniverseCommon(val tablespace: String => Option[String],
 
   val repFor = repForIndex
 
+  val initialLatestDataVersion: Long = 0
```
rjmac: ..which presumably is here?  This isn't used otherwise.

chitang:
migration can be problematic?

rjmac:
Yes, it'll take a lock on an extremely-frequently-used table, so this'll probably have to do the rolling-secondary thing.

rjmac:
Actually it won't take a _read_ lock, so maybe it'll be fine?

PR #304, opened 2019-05-31
EN-32659 Add text to photo conversion

chitang:


chitang:
yes, this is safe to deploy in any order.

PR #305, opened 2019-06-03
EN-32749 Be able to suppress secondary indexing and (re)trigger indexing

chitang:


common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
```diff
@@ -89,10 +98,11 @@ class SecondarySchemaLoader[CT, CV](conn: Connection, dsLogger: Logger[CT, CV],
       val table = tableName(columnInfos)
       val tablespace = tablespaceSqlPart(tablespaceOfTable(table).getOrElse(
         throw new Exception(s"${table} does not exist when creating index.")))
-      using(conn.createStatement()) { stmt =>
+      using(conn.createStatement(), conn.prepareStatement(directivesSql)) { (stmt, directivesStmt) =>
         for {
           (ci, idx) <- columnInfos.zipWithIndex
-          createIndexSql <- repFor(ci).createIndex(table, tablespace)
+          createIndexSql <- repFor(ci).createIndex(table, tablespace) if
+            shouldCreateIndex(directivesStmt, ci.copyInfo.datasetInfo.systemId, ci.fieldName)
```
rjmac: Usually you'd put the `if` on this following line.

common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
```diff
@@ -103,6 +113,34 @@ class SecondarySchemaLoader[CT, CV](conn: Connection, dsLogger: Logger[CT, CV],
...
   }
 
+  /**
+    * To suppress index create
+    *   INSERT into index_directives(dataset_system_id, field_name, directives) VALUES (${datasetSystemId}, ${fieldName}, '{"enable": false}')
+    *      (fieldName can be null to apply to all columns)
+    * To trigger reindex
+    *   curl -v  -X POST -H 'Content-Type: application/json' 'http://localhost:6010/dataset/${resourceName}/!secondary-reindex' --data '{}'
+    */
+  private def shouldCreateIndex(stmt: PreparedStatement, datasetId: DatasetId, fieldName: Option[ColumnName]): Boolean = {
```
rjmac: Hmm.  If I recall correctly the fieldName was optional for migration from a world where field names were not plumbed all the way down to a world where it was.  It might be a good idea to deoptionalize that at some point.  I was confused for a while why this was Optional

common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
```diff
@@ -103,6 +113,34 @@ class SecondarySchemaLoader[CT, CV](conn: Connection, dsLogger: Logger[CT, CV],
...
+    fieldName.flatMap { field =>
+      stmt.setLong(1, datasetId.underlying)
+      stmt.setString(2, field.name)
```
rjmac: Database-level comparisons should be done with `field.casefolded` for case-insensitivity.  There ought to be a column on the table called `field_name_casefolded` or somesuch for that.

common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
chitang: yep, we can do the clean up next

common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
chitang: ok

PR #310, opened 2019-06-27
Changed the expected error code that comes back from invalid GIN inde…

PiSoup:
…xes. Enabled continuation of indexing without forcing a resync


PiSoup:
retest this please

PR #317, opened 2019-11-03
Set the application name when getting a database connection.

znep:
The main use case is for the query server so you can more easily
trace back queries to requests.  This also work in the secondary
watcher code, although is probably less helpful there.

This relies on existing values that are set in the MDC for logging purposes.

Example secondary watcher value: "79 b0a971c5-d3c5-438f-b030-c313f7ee64e8 DatasetId(5)"

Example query server value: "90 4t4jvi0k5qs8qmdxrsygiwr4l _8qy9-my8i"


chitang:
:shipit:

PR #339, opened 2020-03-19
EN-38580: when composing query for text search, attempt to use index

catstavi:
* the to_tsvector search index created on secondaries has the columns ordered by their actual table column names
* however, the search query path works with column aliases, and used to order by those (missing the index)
* this updates the searchVector to attempt an ordering that will hit the index. It uses the underlying column id (ex: '6tcq_mbxm'), plus any subcolumn suffix to determine the order. This is not strictly the true column name (ex: u_6tcq_mbxm_28) but will work for ordering.
* if not all the CoreExpr in the searchVector are ColumnRefs, its not going to be possible for us to use any index anyway.

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/store/index/FullTextSearch.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/store/index/FullTextSearch.scala
@@ -1,11 +1,12 @@
 package com.socrata.pg.store.index
 
+import com.socrata.datacoordinator.id.UserColumnId
 import com.socrata.datacoordinator.truth.sql.SqlColumnCommonRep
 import com.socrata.pg.soql.Sqlizer._
 import com.socrata.pg.soql.SqlizerContext._
 import com.socrata.soql.collection.OrderedMap
 import com.socrata.soql.environment.{ColumnName, TableName}
-import com.socrata.soql.typed.CoreExpr
+import com.socrata.soql.typed.{ColumnRef, CoreExpr}
 import com.socrata.soql.types.SoQLUrl
 
 trait FullTextSearch[CT] {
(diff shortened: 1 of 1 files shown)


common-pg/src/main/scala/com/socrata/pg/store/index/FullTextSearch.scala
```diff
@@ -23,19 +24,29 @@ trait FullTextSearch[CT] {
     val phyCols = repsSearchableTypes.flatMap(rep => rep.physColumns.map(phyCol =>
       coalesce(ctx.map(qualify(phyCol, _)).getOrElse(phyCol))
     ))
-    toTsVector(phyCols)
+    toTsVector(phyCols.zip(phyCols))
   }
 
-  def searchVector(selection: OrderedMap[ColumnName, CoreExpr[_, CT]], ctx: Option[Context]): Option[String] = {
-    val cols = selection.view.filter(x => SearchableTypes.contains(x._2.typ)).flatMap { case (columnName, expr) =>
+  def searchVector(selection: OrderedMap[ColumnName, CoreExpr[UserColumnId, CT]], ctx: Option[Context]): Option[String] = {
```
PiSoup: Was this type previously anonymous for a specific reason?

catstavi: 🤔 I don't know, but I thought the compiler would tell me if there were unsafe uses of it. I could be cautious and leave it anonymous so if there are any that don't match, it just goes to the case statement default.

rjmac: It just didn't care about that type before

PiSoup: With type erasure I don't think the compiler would be able to complain in this case. Don't know for sure though.

catstavi: removing the type here gives a type erasure error on the case match line. I'm learning more about it, but @rjmac you're saying its ok to leave the type match in the function head here?

rjmac: Scala doesn't do matching in function heads.  So, yes.  Anything that called it with the wrong type would complain.

PR #343, opened 2020-04-08
Chi/en 39051/param misalign

chitang:


chitang:
oops, not correct

chitang:
Take 2.  I think this makes more sense.

common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
```diff
@@ -82,9 +82,14 @@ object SoQLAnalysisSqlizer extends Sqlizer[AnalysisTarget] {
...
-          rep, typeRep, subParametricSql.setParams, subCtx, escape)
-      (sqls, ana)
+      val (sqls, paramsCountInSelect) = sql(ana, Some(prevAna), tableNamesSubTableNameReplace, allColumnReps, reqRowCount &&  ana == lastAna,
+          rep, typeRep, Seq.empty, subCtx, escape)
+
+      // query parameters in the select phrase come before those in the sub-query.
+      // query parameters in the other phrases - where, group by, order by, etc come after those in the sub-query.
+      val (setParamsBefore, setParamsAfter) = sqls.setParams.splitAt(paramsCountInSelect)
+      val setParamsAcc = setParamsBefore ++ subParametricSql.setParams ++ setParamsAfter
+      (sqls.copy(setParams = setParamsAcc), ana)
```
rjmac, line 92: This makes my skin crawl (it absolutely screams "sql generation is using the wrong abstraction"), but no more than anything else in here to do with joins...

PR #345, opened 2020-06-03
EN-40387: Not selecting group by constant result in error

chitang:
QA: To repro:
1. create soql view - select field, ' ' as constant group by field, constant
2. $query=select field -> error

### DIFF
--- a/common-pg/src/test/scala/com/socrata/pg/soql/SqlizerBasicTest.scala
+++ b/common-pg/src/test/scala/com/socrata/pg/soql/SqlizerBasicTest.scala
@@ -395,17 +395,20 @@ class SqlizerBasicTest extends SqlizerTest {
     params should be(Seq("oNe"))
   }
 
-  test("group by literals") {
+  test("group by literals with constants removed") {
     val soql = "select id, 'stRing' as a, 5 as b, 2*3 as c group by id, a, b, c"
     val ParametricSql(Seq(sql), setParams) = sqlize(soql, CaseSensitive)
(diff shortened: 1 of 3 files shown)


common-pg/src/test/scala/com/socrata/pg/soql/SqlizerBasicTest.scala
```diff
@@ -395,10 +395,17 @@ class SqlizerBasicTest extends SqlizerTest {
...
     val ParametricSql(Seq(sql), setParams) = sqlize(soql, CaseSensitive)
-    sql should be ("SELECT t1.id,e'stRing',5,(2 * 3) FROM t1 GROUP BY t1.id,2,3,4")
+    sql should be ("SELECT t1.id,e'stRing',5,(2 * 3) FROM t1 GROUP BY t1.id")
+    setParams.length should be (0)
+  }
+
+  test("group by literals all removed") {
+    val soql = "select 'stRing' as a, 5 as b, 2*3 as c group by a, b, c"
+    val ParametricSql(Seq(sql), setParams) = sqlize(soql, CaseSensitive)
+    sql should be ("SELECT e'stRing',5,(2 * 3) FROM t1")
```
znep, line 408: Wouldn't this need a `limit 1` to be equiv?

chitang: hm...yes, this edge case wouldn't work.

chitang: fixed in a different way

PR #360, opened 2020-09-15
EN-24494: (leading) search sql generation bug around previous analysis

chitang:
Cheetah regression:
querying a view with search
  should honor search saved in a derived view with soql using soda v2.1 from nbe


vickingan (approved):
👍

PR #366, opened 2020-10-06
Do not refresh rollups when rows are empty

chitang:


store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -396,7 +396,7 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
         case RowDataUpdated(ops) =>
           val loader = dataLoader.getOrElse(prevettedLoader(pgu, truthCopyInfo))
           RowDataUpdatedHandler(loader, ops)
-          (rebuildIndex, true, truthCopyInfo, Some(loader))
+          (rebuildIndex, refreshRollup || ops.headOption.isDefined, truthCopyInfo, Some(loader))
```
rjmac, line 399: `ops.nonEmpty` should work

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -396,7 +396,7 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
         case LastModifiedChanged(lastModified) =>
           pgu.datasetMapWriter.updateLastModified(truthCopyInfo, lastModified)
           (rebuildIndex, true, truthCopyInfo, dataLoader)
```
rjmac, line 402: This will happen on every single update a secondary gets, so the `true` here should probably be `refreshRollup` too.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -362,12 +362,12 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
           FieldNameUpdatedHandler(pgu, truthCopyInfo, secondaryColInfo)
           (rebuildIndex, false,  truthCopyInfo, None)
```
rjmac, line 363: This should also probably be `refreshRollup`.  It won't happen in changesets generated via the public API, but it's totally possible to tell data coordinator "here are a bunch of rows, oh and change that fieldname too" in a single mutation script.

rjmac (changes requested):
Generally looks good, just a couple of details.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
rjmac: ..actually thinking about it, this change is the _only_ important one.  An empty changeset will produce _only_ a LastModifiedChange event, no rows at all (though the other change, or-ing in `nonEmpty` makes sense and I think it should happen)

rjmac (approved):

PR #368, opened 2020-10-21
EN-42578: Plug rollup table leak

rjmac:
The code assumed that data versions would increase by 1 each time; this
is no longer true.  Instead, fetch the previous data version from the
database.

Also remove the 0-row optimization, because it will break things (the
data version will still be changed, and therefore the table name)


chitang:
:shipit: , is it worth the effort to fix and keep the rollup skip update optimization?

rjmac:
Hm, RollupDropTest passed locally.  Digging into that...

Also, I'm not sure.  Because the rollup table name includes the data version, _something_ will have to change (even a 0-row update bumps the data version) but just renaming the table is fraught with problems because of locks, so we'd have to create a new one and populate it from the old, which is possible but a whole new code path, so.. maybe not.

rjmac:
Oh, now I understand why that change fixes the test.  Sigh, universal equality :typeerror:

PR #376, opened 2021-01-26
EN-43695: Add a `get_context` function and allow clients to send context info

rjmac:


soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
```diff
@@ -116,13 +117,37 @@ trait DataSqlizerQuerier[CT, CV] extends AbstractRepBasedDataSqlizer[CT, CV] {
...
+      // works fine for the variable name, I'm md5ing it because I'm
+      // not willing to expose our databases to potential hostile
+      // activity in a thing that's not generally considered part of
+      // the query path.
+      //
+      // Because these things are mainly/exclusively set by core, I'm
+      // not going to care overmuch about the values' trustworthiness.
+      // This is mostly to prevent hypothetical attacks via the
+      // get_context function.
+      using(conn.prepareStatement("select set_config('user_variables.' || md5(?), ?, true)")) { stmt =>
```
chitang, line 132: not sure if it is worthwhile to group that into batch of 5 so it can set multiples in one trip.

chitang (approved):

soql-server-pg/src/main/scala/com/socrata/pg/query/DataSqlizerQuerier.scala
rjmac: It would not be difficult to do, so therefore I shall do it!  (Though I'll be surprised if we ever send more than one or two in practice)

PR #378, opened 2021-02-08
Union - BinaryTree

chitang:
I have not published soql reference yet.  It can't build.  But i want a quick glance before we do.

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
@@ -1,142 +1,282 @@
 package com.socrata.pg.soql
 
-import com.socrata.NonEmptySeq
 import com.socrata.datacoordinator.id.UserColumnId
 import com.socrata.datacoordinator.truth.sql.SqlColumnRep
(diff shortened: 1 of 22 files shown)


common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
```diff
@@ -1,142 +1,282 @@
...
+          // simple join w/o alias does not need remap
+          acc + (name -> name)
+        case _ =>
+          // complex join reps are just qualifier.name and use type reps and don't need entries in reps.  They use
+          acc
+      }
+    }
+
+    val ctx = context + (Analysis -> analysis) +
+      (SimpleJoinMap -> simpleJoinMap) +
```
vickingan, line 207: Are we adding all the forms of joins for VQE here?

chitang: There are two types of joins - complex like "join (select...)" and simple join view.  We only need a map of simple joins.

vickingan (approved):
LGTM though may also want to add a reviewer who knows more about this part of the code

PR #380, opened 2021-02-12
EN-44281: Union/Query BinaryTree Rollup not used fix

chitang:
1. use rollup table name
2. use type rep if rep is not found


soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
```diff
@@ -435,6 +434,13 @@ class QueryServer(val dsInfo: DSInfo, val caseSensitivity: CaseSensitivity, val
           acc ++ getJoinReps(pgu, copyInfo, tableName)
         }
 
+        val tableName = rollupName match {
+          case Some(_) =>
+            querier.sqlizer.dataTableName
```
chitang, line 439: let me try to just always use querier.sqlizer.dataTableName

rjmac (approved):

PR #382, opened 2021-02-22
EN-44488: Request rows_count fails for some queries

chitang:


common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
```diff
@@ -165,7 +165,9 @@ trait SoQLAnalysisSqlizer {
     // TODO: switch to trailing search when there is smarter index support
     val leadingSearch = context.getOrElse(LeadingSearch, true) != false
 
-    val ana = if (reqRowCount) rowCountAnalysis(analysis) else analysis
+    val isOutermost = isOutermostAnalysis(analysis, context)
+    val reqRowCountFinal = reqRowCount && isOutermost
+    val ana = if (reqRowCountFinal) rowCountAnalysis(analysis) else analysis
```
langlk, line 170: Okay, let me make sure I understand this:

- If we `reqRowCount` is true, and this is the outermost analysis, we get the row count.
- Otherwise we do not

Is that correct?

chitang: yes, the row count special treatment

langlk (approved):
LGTM!

PR #383, opened 2021-03-04
Add timing index creation

chitang:


common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryUniverse.scala
```diff
@@ -73,7 +73,7 @@ class PGSecondaryUniverse[SoQLType, SoQLValue](
   }
 
   def schemaLoader(logger: Logger[SoQLType, SoQLValue]): SecondarySchemaLoader[SoQLType,SoQLValue] =
-    new SecondarySchemaLoader(conn, logger, repForIndex, tablespace, commonSupport, RowSizeBufferSqlErrorContinue)
+    new SecondarySchemaLoader(conn, logger, repForIndex, tablespace, commonSupport, RowSizeBufferSqlErrorContinue, timingReport)
```
vickingan, line 76: Where does the variable `timingReport` come from in this file?

vickingan (approved):
👍

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryUniverse.scala
chitang: timingReport came from the caller PGSecondaryUniverse

PR #390, opened 2021-04-06
Translate SoQL minus to SQL except

chitang:


rjmac (approved):
:+1: though I'd prefer it if we weren't dumping text received in the analysis into PG, even if the analysis is received from a trusted source... it'd be nicer if soqlQueryOp were some kind of enum.

PR #391, opened 2021-04-13
EN-45573 SoQL Json

rozap:
initial soql json support


rozap:
closed in favor of #393

PR #392, opened 2021-04-14
Make the decision about whether to copy on upsert more flexible

rjmac:
Also configure it in a less-silly way.

### DIFF
--- a/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
+++ b/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
@@ -53,14 +53,8 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
     client
   }
 
-  val rowsChangedPreviewConfig = curator match {
-    case Some(c) =>
-      val config = new RowsChangedPreviewConfig.ZKClojure(c, "/soql-postgres-secondary")
-      config.start()
-      config
-    case None =>
-      RowsChangedPreviewConfig.Default
-  }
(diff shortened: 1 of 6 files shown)


rjmac:
..huh, that wasn't a build failure I expected.

rjmac:
Oh, that's a silly mistake in a thing I adjusted at the last minute, of course.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -53,14 +53,8 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
-  val rowsChangedPreviewConfig = curator match {
-    case Some(c) =>
-      val config = new RowsChangedPreviewConfig.ZKClojure(c, "/soql-postgres-secondary")
-      config.start()
-      config
-    case None =>
-      RowsChangedPreviewConfig.Default
-  }
+  val rowsChangedPreviewConfig =
+    new RowsChangedPreviewConfig.Clojure(storeConfig.transparentCopyFunction.getOrElse(RowsChangedPreviewConfig.default))
```
vickingan, line 57: Not a clojure expert, but `transparentCopyFunction` and `default` look like same function?

rjmac: They do and are!  But we can change transparentCopyFunction in apps_marathon.

vickingan (approved):
👍 LGTM

PR #393, opened 2021-04-14
EN-45612 Json casts 

rozap:
* Add json cases
* Add json subscript functions
* to be merged after https://github.com/socrata-platform/soql-reference/pull/183


common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -153,6 +153,9 @@ object SqlFunctions extends SqlFunctionsLocation with SqlFunctionsGeometry with
 
     TextToBool -> formatCall("%s::boolean") _,
     BoolToText -> formatCall("%s::varchar") _,
+    TextToJson -> formatCall("%s::jsonb") _,
+    JsonToText -> formatCall("%s::text") _,
+    JsonToNumber -> formatCall("%s::numeric") _,
```
rjmac, line 158: JsonToText and JsonToNumber do fundamentally different things -- the former returns the JSON serialized to text, the latter converts a json number to a number and raises an error if it can't.  I think the latter should be a function not a cast (and, if there's an extract-a-number function, there should probably also be extract-a-string and extract-a-boolean too).

Also right now if texttojson or jsontonumber fail, they'll explode in a glorious 500 postgresql error.  Which is our normal reaction to domain errors, alas, but is't great.

rjmac: ..oh actually that's not even true, about JsonToNumber.  Casting a jsonb to numeric is just a type error in postgresql full stop.

rozap: yea, i think that's fair. this cast could use some more thought and likely should be a function. it's not needed for the basic functionality so i think i'll just remove it.

rjmac (approved):

PR #394, opened 2021-04-26
Handle the single-row table special case

rjmac:
* Create a columnless `onerow` table we can reference
* Only actually generate that reference if necessary


znep:
Is there a reason to not call the table in postgres `single_row` too?

rjmac:
Nope, I can do that.

vickingan (approved):
👍

PR #395, opened 2021-05-17
add regr_slope/r2/intercept and natural log functions

PeterAustinMoore:
Added functions for:
`regr_slope`
`regr_intercept`
`regr_r2`
`ln`


chitang (commented):
need to bump soql-reference too

chitang (approved):

PR #396, opened 2021-05-18
Eliminate SQL references to @single_row in more cases

rjmac:


common-pg/src/main/scala/com/socrata/pg/soql/SoQLAnalysisSqlizer.scala
```diff
@@ -461,7 +474,18 @@ trait SoQLAnalysisSqlizer {
...
+                               setParams ++
+                                 setParamsSelect ++
+                                 finalSetParamsJoin ++
+                                 setParamsWhere ++
+                                 setParamsWhereSearch ++
+                                 setParamsGroupBy ++
+                                 setParamsHaving ++
+                                 setParamsHavingSearch ++
+                                 setParamsOrderBy ++
+                                 setParamsOrderBy)
```
chitang, line 487: setParamsOrderBy added twice

rjmac: Yeah, that's one of the test failures (two of the test failures, actually).  The others are in soql-reference so I'll be making a new PR there too.

chitang (approved):

PR #401, opened 2021-06-04
Remove remnants of previous window function implementation

chitang:


saraid (approved):
I don’t have the background to verify that this won’t break anything and/or is complete, but it looks fine.

PR #402, opened 2021-06-28
Be less eager to recreate the full text search index for datasets

rjmac:
In particular, only do it if its definition has actually changed.

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
@@ -60,9 +60,9 @@ class SecondarySchemaLoader[CT, CV](conn: Connection, dsLogger: Logger[CT, CV],
     // index.
     //
     // Postgresql does give a way to get SQL for the index in the
-    // pg_indexes table, but it's beenñ normalized and it would
-    // be silly to fight with PG implementation details to figure out
-    // if our sql is identical to that SQL.
+    // pg_indexes table, but it's been normalized and it would be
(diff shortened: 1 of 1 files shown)


common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
```diff
@@ -48,22 +50,61 @@ class SecondarySchemaLoader[CT, CV](conn: Connection, dsLogger: Logger[CT, CV],
...
+    // it isn't already what it's supposed to be, and "what it's
+    // supposed to be" is more complicated than for individual column
+    // indices (e.g., if a new column is created, this index needs to
+    // be recreated if it already exists).  In order to find out if we
+    // need to do this, we'll use postgresql's comment facility to tag
+    // the index with a hash of the SQL we're using to define the
+    // index.
+    //
+    // Postgresql does give a way to get SQL for the index in the
+    // pg_indexes table, but it's beenñ normalized and it would
```
saraid, line 63: `beenñ`

saraid (approved):
This makes sense to me. Weird, but makes sense.

common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
rjmac: Isn't that how one conjugatedñ english past-tense verbs?

saraid: No it isn'tñ.

chitang:
:shipit:

PR #403, opened 2021-06-29
EN-46905: Add new soql functions

chitang:


saraid (approved):
I can't follow what `nary` is doing, but I assume it's fine and appropriate for this. Slightly baffled as to why it's not appropriate for split_part, though, in that case.

chitang:
both nary and formatcall should work.  formatcall seems be more direct and easier to read in that case where there is fixed number of parameters.

PR #405, opened 2021-07-06
EN-46905: Add more soql string functions

chitang:


chitang:
Will bump to the correct soql version after you approve soql-reference.

langlk (approved):

PR #420, opened 2021-11-05
Add interval type and date_add function

chitang:


svjayashree:
@chitang - if we are adding new soql functions, going forward, the requirement is that we should add the documentation for it as well. Can you take a stab at adding the documentation for date_add please?

chitang:
There is documentation which lives in soql-reference, not this project.

common-pg/src/test/scala/com/socrata/pg/store/PGSecondaryUniverseTestBase.scala
```diff
@@ -59,7 +59,8 @@ trait PGSecondaryUniverseTestBase extends FunSuiteLike with Matchers with Before
     // Setup the data columns
     val cols = SoQLType.typesByName.filterKeys(t => !UnsupportedTypes.contains(t) &&
                                                     t != SoQLID.name && // These types are specially added.
-                                                    t != SoQLVersion.name)
+                                                    t != SoQLVersion.name &&
+                                                    t != SoQLInterval.name)
```
catstavi, line 63: It looks like there are two groups of types here -- "UsupportedTypes" (json) which seems to have aspirations as eventually being a dataset column type? and then SoQLID and SoQLVersion which are special for other reasons?

Would it be right to group SoQLInterval with UsupportedTypes instead of the id/version?

catstavi: oh, you did add it to UnsupportedTypes! does it need to be added here as well then?

chitang: We don't need it.  Removed.

catstavi (approved):

PR #424, opened 2021-11-12
EN-49155: Support query either via url param or body

chitang:


chitang:
Take 2

soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
```diff
@@ -243,7 +244,14 @@ class QueryServer(val dsInfo: DSInfo, val caseSensitivity: CaseSensitivity, val
   private def query(explain: Boolean)(req: HttpRequest)(resp: HttpServletResponse): Unit =  {
     val servReq = req.servletRequest
     val datasetId = servReq.getParameter("dataset")
-    val analysisParam = servReq.getParameter("query")
+    val analysisParam = Option(servReq.getParameter("query")) match {
+      case Some(q) => q
+      case _ =>
+        using(servReq.getInputStream) { ins =>
```
sjzavaleta, line 250: What else might be inside the inputStream? Nothing? Who decides to either put a query either in the parameter query or in the inputSteam (= body?)?

chitang: The body only contains the query analysis.  QC is the only client and decide where to put it.

saraid (approved):
Whatever.

PR #433, opened 2021-12-02
EN-49288: Remove ::number for count(*) at the wrong position

chitang:


common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -538,7 +538,7 @@ object SqlFunctions extends SqlFunctionsLocation with SqlFunctionsGeometry with
                  ctx: Sqlizer.Context,
                  escape: Escape): ParametricSql = {
 
-    val templateTypeCast = if (fn.window.isEmpty && typeCastIfNotWindowFn.isDefined) template + typeCastIfNotWindowFn.map("::" + _).get
+    val templateTypeCast = if (fn.window.isEmpty && fn.filter.isEmpty && typeCastIfNotWindowFn.isDefined) template + typeCastIfNotWindowFn.map("::" + _).get
```
catstavi, line 541: so this is now like... typecast if not window or filter function?

chitang: type if (not window) and (not filter) and ...

chitang: i was handling count(column) but missed count(*) because count_star is a weird function that is handled differently.

catstavi: just the naming has become misleading is what I'm observing: `typeCastIfNotWindowFn`

catstavi: But I don't know what a better name would be without getting really silly-long

catstavi (approved):
:( sad approve because this fix should go out but I wish for name improvements

common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
chitang: That part isn't changed although the name is kinda tied.  I can rename that to typeCastIfNotWindowFnAndNotFilter or simply typeCast.

catstavi (approved):
not sad approve! thank you!

PR #439, opened 2021-12-30
EN-49831: Support SoQL Hint

chitang:
SELECT col1, col2 WHERE . . . |> 
  SELECT HINT(materialized) col1


saraid (approved):
I don't actually understand what's going on, but fine whatever.

PR #440, opened 2022-01-05
Streamline datetime literal parameters

chitang:
Previously going through cstring:
Filter: ((u_4n3z_8mii_6 >= ('2020-02-11T00:00:00.000'::cstring)::timestamp without time zone)

New:
Filter: (u_xh59_u4rh_18 = '2016-01-01 03:00:00'::timestamp without time zone)

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
@@ -10,7 +10,6 @@ import com.socrata.soql.types.SoQLVersion.{StringRep => SoQLVersionRep}
 import com.socrata.soql.types._
 import Sqlizer._
 import SoQLFunctions._
-import com.socrata.pg.soql.SqlizerContext.{RootExpr, SoqlPart}
 import com.socrata.soql.exceptions.BadParse
 import org.joda.time.format.{PeriodFormatter, PeriodFormatterBuilder}
 import org.joda.time.{DateTime, LocalDateTime, Period}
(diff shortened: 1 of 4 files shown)


common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
```diff
@@ -620,8 +619,14 @@ object SqlFunctions extends SqlFunctionsLocation with SqlFunctionsGeometry with
...
 
+    val litCtx = soqlType match {
+      case SoQLFloatingTimestamp | SoQLFixedTimestamp =>
+        ctx + (SqlizerContext.TimestampLiteral -> true)
+      case _ =>
+        ctx
+    }
     val fnWithConvertedParams = fn.copy(parameters = convertedParams, window = fn.window)
-    formatCall(template, None, None)(fnWithConvertedParams, rep, typeRep, setParams, ctx, escape)
+    formatCall(template, None, None)(fnWithConvertedParams, rep, typeRep, setParams, litCtx, escape)
```
rjmac, line 629: It feels like it'd be cleaner just to do it here?

```scala
match (soqlType, convertedParams) {
  case (SoQLFloatingTimestamp, StringLiteral(SoQLFloatingTimestamp.StringRep(ts), _)) =>
    // at this point ts is a LocalDateTime and we can generate a ParametricSql that is just that
  case (SoQLFixedTimestamp, StringLiteral(SoQLFixedTimstamp.StringRep(ts), _)) =>
    // at this point ts is a DateTime and we can generate a ParametricSql that is just that
  case _ =>
      ... old codegen here ...
}
```

chitang: Can't share much of the crazy code (like logic to not use parameter if it is in group by) if it is done here.

catstavi:
looked through it and I'm good when rjmac is good 👍

common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctions.scala
chitang: like this:
select count(id),
round((extract(epoch from created_at) - extract(epoch from '2012-01-01'::timestamp with time zone))::numeric/86400)
from lenses
group by round((extract(epoch from created_at) - extract(epoch from '2012-01-01'::timestamp with time zone))::numeric/86400)

rjmac:
> looked through it and I'm good when rjmac is good +1

We talked in Slack about that approach, and I'm good too :+1:

rjmac:
..huh interesting, emojis get de-emoji'd when you quote reply!

catstavi (approved):
yay performance!

PR #441, opened 2022-01-06
Fix excessive setParam calls

chitang:


version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "4.0.31-SNAPSHOT"
+ThisBuild / version := "4.1.0-SNAPSHOT"
```
sjzavaleta, line 1: is this related?

chitang: No, but I missed the opportunity to ask for a bigger version change lately.  We don't really strictly follow semantic versioning.  There are changes that are significant enough to have a bigger version bump.

chitang: We should have a major bump everytime we change analysis (de)serialization.

sjzavaleta (approved):

PR #442, opened 2022-01-20
reproject soql function

PeterAustinMoore:
Adds:
A `reproject` SoQL function that loosely does what ST_Transform does in PostGIS, however it is setup to accept only the Proj4 projection text, rather than SRID integers, in order to allow for greater flexibility as well as match what DSMAPI does as a transform: https://dev.socrata.com/docs/transforms/reproject.html

Example: 
```
reproject('POINT (372728.3308536674 69825.24782297359)'::point, '+init=epsg:3627', '+init=epsg:4326')
```

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -19,7 +19,7 @@ object Dependencies {
     val socrataCuratorUtils = "1.2.0"
     val socrataThirdPartyUtils = "5.0.0"
     val socrataHttpCuratorBroker = "3.13.4"
-    val soqlStdlib = "4.0.1"
+    val soqlStdlib = "4.0.3"
     val typesafeConfig = "1.0.0"
     val dataCoordinator = "3.8.12"
     val typesafeScalaLogging = "3.9.2"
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -19,7 +19,7 @@ object Dependencies {
     val socrataCuratorUtils = "1.2.0"
     val socrataThirdPartyUtils = "5.0.0"
     val socrataHttpCuratorBroker = "3.13.4"
-    val soqlStdlib = "4.0.1"
+    val soqlStdlib = "4.0.3"
```
catstavi, line 22: have you verified with author of 4.0.2 that it's ok to bump this? (i see no reason not, but just to be safe!)

PeterAustinMoore: Yup! Chatted with Chi and he said it is good to go

PeterAustinMoore:
We decided this would not be performant for many operations, encourages bad practices (i.e. not transforming it on ingress), and might cause CPU issues that would negatively affect everyone.

PR #446, opened 2022-02-05
EN-50639: Do not rebuild all existing rollups from scratch when one is added or deleted.

chitang:


chitang:
I plan to look at possible lock/timeout and might fall back to the old style or a copy.

common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
```diff
@@ -140,6 +140,22 @@ class RollupManager(pgu: PGSecondaryUniverse[SoQLType, SoQLValue], copyInfo: Cop
...
   }
 
+  def bumpVersionRollup(rollupInfo: RollupInfo, oldCopyInfo: Option[CopyInfo], newDataVersion: Long): Boolean = {
+    logger.info(s"Updating copy $copyInfo, rollup $rollupInfo")
+    time("update-rollup", "dataset_id" -> copyInfo.datasetInfo.systemId, "rollupName" -> rollupInfo.name) {
+      val newTableName = rollupTableName(rollupInfo, newDataVersion)
+      oldCopyInfo.filter(_.copyNumber == rollupInfo.copyInfo.copyNumber).foreach { ci =>
+        val oldTableName = rollupTableName(rollupInfo, ci.dataVersion)
+        using(pgu.conn.createStatement()) { stmt =>
+          val sql = s"ALTER TABLE ${oldTableName} RENAME TO ${newTableName}"
```
znep, line 150: Is this going to cause locking issues for any queries trying to use this table?  My understanding was renaming a table required an exclusive lock.   eg. if you add a new rollup that takes a while to build, and have already renamed another rollup in that transaction, won't that block read queries that try to access it?

chitang: I combine all alter table statements into one and give it 60s timeout.  Other long but safe operations are all executed before alter tables.  If alter tables does time out due to lock, it will roll back and fall back to the slow method.
The slow method drops rollup tables immediately.  I wonder why the slow method would not cause similar exclusive lock issue (I haven't verified about that yet.)

znep: I'm surprised the existing method drops rollup tables immediately.

I still have concerns about what happens to queries reading from the rollups while this happens.

chitang: Do you have a concrete example that will cause problem with the logic?

znep: 1. have a query reading from the rollup, which takes a shared lock
2. have an alter blocking on a lock (or anything else after that but before commit)
3. when a second query tries to read from the rollup, it will now block trying to get a shared lock

chitang: My puzzle about the existing immediate drop is solved.  It is trying to drop new version of rollup before creating the new version.  It isn't dropping the old version.

About your example, yes - it will block other read queries but only up to 60s and then it will fall back to the slow method.  I am okay to lower that number or better yet, make that configurable.

rjmac: Perhaps an alternative to renaming is `create table the_new_name like the_old_name; insert into the_new_name select * from the_old_name` would work?  It would copy all the data, but it wouldn't need to _recompute_ all the data...

znep: @rjmac that would definitely help for the initially designed use case where rollups are small, but I worry that some of our current cases have rather large "rollups" that aren't really doing much aggregation of data so would still be almost as slow.  We may want to dig into if that worry is justified privately.

Oh, it would also need to create indexes too, as rollups currently get the full index treatment in an attempt to ensure you don't end up with queries being slower hitting an unindexed rollup.

znep: Another option is changing the naming scheme so there is a level of indirection so we don't have to have the rollup tables named for the data version, so they don't have to be renamed/recreated unless actual data changes, but I haven't thought that through.

chitang: Even for < 10s?  This is especially useful for development I wish we have a mechanism as flexible as soql hints.

chitang: I thought about the create table like and insert into.  That will be a half way compromise.

chitang: how about a 10s timeout and falling back to create table like?

znep: I am not a big fan of adding multiple hard to tests codepaths.  It has repeatedly bit us in related areas in the past.

chitang: I think it is well worth the speed gain.  But I am not insisting and remove the speediest path for now.

common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
```diff
@@ -140,6 +140,19 @@ class RollupManager(pgu: PGSecondaryUniverse[SoQLType, SoQLValue], copyInfo: Cop
     }
   }
 
+  def bumpRollupVersionByCopyTable(rollupInfo: RollupInfo, oldCopyInfo: Option[CopyInfo], newDataVersion: Long): Unit = {
+    val newTableName = rollupTableName(rollupInfo, newDataVersion)
+    oldCopyInfo.filter(_.copyNumber == rollupInfo.copyInfo.copyNumber).map { ci =>
+      val oldTableName = rollupTableName(rollupInfo, ci.dataVersion)
+      using(pgu.conn.createStatement()) { stmt =>
+        val sql = s"CREATE TABLE ${newTableName} (LIKE ${oldTableName} including all); INSERT INTO ${newTableName} SELECT * FROM ${oldTableName};"
+        logger.info(sql)
```
znep, line 149: Can you either remove this log or add some text describing it (eg. "Copying rollup using sql: " ...), to avoid future someone having a hard time knowing where this log message comes from?

common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
```diff
@@ -140,6 +140,19 @@ class RollupManager(pgu: PGSecondaryUniverse[SoQLType, SoQLValue], copyInfo: Cop
     }
   }
 
+  def bumpRollupVersionByCopyTable(rollupInfo: RollupInfo, oldCopyInfo: Option[CopyInfo], newDataVersion: Long): Unit = {
+    val newTableName = rollupTableName(rollupInfo, newDataVersion)
+    oldCopyInfo.filter(_.copyNumber == rollupInfo.copyInfo.copyNumber).map { ci =>
+      val oldTableName = rollupTableName(rollupInfo, ci.dataVersion)
+      using(pgu.conn.createStatement()) { stmt =>
+        val sql = s"CREATE TABLE ${newTableName} (LIKE ${oldTableName} including all); INSERT INTO ${newTableName} SELECT * FROM ${oldTableName};"
```
znep, line 148: I am a bit concerned this may be slower for large rollups as it inserts data into a table that is already indexed instead of inserting the data then creating indexes.  In the past, what I've seen is that for large rollups inserting the data and creating the indexes are 90%+ of the time and the actual query is <10%.

Have you looked at how hard it would be to create it without indexes then run the codepath that creates the indexes?  I'm not sure how flexible that code is.

This codepath only gets hit for rollup creation / deletion without data changes, so maybe that is an ok tradeoff for now?  At the least it would be good to have a comment here documenting that tradeoff.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -645,14 +657,81 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
+      updateRollupsSlow(pgu, originalCopyInfo, copyInfo, rollupDirty)
+    } else {
+      updateRollupsFast(pgu, originalCopyInfo, copyInfo, rollupDirty),
+    }
+  }
+
+  private def updateRollupsSlow(pgu: PGSecondaryUniverse[SoQLType, SoQLValue],
+                                originalCopyInfo: Option[TruthCopyInfo],
+                                copyInfo: TruthCopyInfo,
+                                rollupDirty: RollupDirty): Unit = {
```
znep, line 672: `rollupDirty` is unused.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -645,14 +657,81 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
     pgu.datasetMapReader.rollups(copyInfo).foreach { ri =>
       rm.updateRollup(ri, originalCopyInfo, copyInfo.dataVersion)
     }
   }
 
+  /**
+   * TODO: Handle each changed rollup separately fallback each blocked one to copy table
+   *       without switching back to the slow path for the whole batch.
+   */
+  private def updateRollupsFast(pgu: PGSecondaryUniverse[SoQLType, SoQLValue],
```
znep, line 685: Can we add a comment explaining in words the logic in `updateRollupsFast`?

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -645,14 +657,81 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
+        rollups.get(ri.name) match {
+          case Some(r) =>
+            rm.dropRollup(r, true)
+            rm.updateRollup(r, originalCopyInfo, copyInfo.dataVersion)
+          case None =>
+            logger.warn(s"cannot find rollup ${ri.name} - ${copyInfo.datasetInfo.systemId}")
+        }
+      }
+
+      rollupDirty.delete.foreach { ri =>
```
znep, line 720: Is it possible to end up with both delete and update operations in one version?  In that case it seems we would need more logic to avoid removing rollups that where removed then updated (ie. created).  I'm not sure if the version combining code could result in that.

common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
znep: Also, note that this will result with indexes being named differently on the rollup.  I don't think that anything cares about that but wanted to call it out.

I also did a test just to satisfy myself there are no exclusive locks required on the source table.  I know there _shouldn't_ logically be but have run into somewhat related things in the past so wanted to be sure.

session 1:

```
[local]:5432 marcslemko@marcslemko=# begin;
BEGIN
[local]:5432 marcslemko@marcslemko=#* select * from foo;
 bar 
-----
   1
(1 row)

[local]:5432 marcslemko@marcslemko=#* \d foo
                Table "public.foo"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 bar    | integer |           |          | 
Indexes:
    "marcs1" btree (bar)
    "marcs2" btree (bar)
```
(transaction left open)

session 2:

```
[local]:5432 marcslemko@marcslemko=# begin;
BEGIN
[local]:5432 marcslemko@marcslemko=#* create table foo2 (like foo including all);
CREATE TABLE
[local]:5432 marcslemko@marcslemko=#* insert into foo2 select * from foo;
INSERT 0 1
[local]:5432 marcslemko@marcslemko=#* commit;
COMMIT
[local]:5432 marcslemko@marcslemko=# \d foo2
                Table "public.foo2"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 bar    | integer |           |          | 
Indexes:
    "foo2_bar_idx" btree (bar)
    "foo2_bar_idx1" btree (bar)
```

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
chitang: I am aware but am trying to keep the signature identical.  Well...I don't insist.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
chitang: good call...should just assume it can happen and handle it.

chitang:
This is my summary.  I think we should do these:
Add truth.rollup.version, rollup_map.version to help generate more stable rollup names. EN-50425
Combine consecutive rollup versions when playing back versions EN-50425
Add rollup capability in dsmapi so that it can avoid full replace. EN-50485

I am staying out of all three and let Robert take on them.  I am focusing on the WHERE support work.

PR #448, opened 2022-02-09
NOJIRA: Remove unnecessary non-alpha pg secondary solo services

rlvoyer:
Are these useful for local development?


rlvoyer:
`solo register --all` does what it says and ignores the `register_by_default` property, which is already set to `false` in the service definition for the services in question. If I register w/o using the `--all` flag, I get the desired behavior.

PR #450, opened 2022-02-17
Add data shape version

rjmac:
This doesn't actually care about it, but because it uses DC's database
code it needs to fake caring.


chitang:
I thought we are going to actually make use of data_shape_version for rollup names to significantly improve rollup write efficiency.

sjzavaleta (approved):

rjmac:
I didn't think that; that was what I did the batching stuff for.

rjmac:
(we can always add DSV to secondaries if we decide that we do need it, but it requires a change to the secondary API)

chitang:
For example, a dataset is published, has 69 rollups and fully replicated.  Then, add one more rollup.  Currently, it will have to rewrite all 69 + 1 rollups.  I think we can do much better and just refresh one rollup.

PR #459, opened 2022-04-11
EN-49039: Limit $$query_timeout_seconds to 10 minutes unless it is debug

chitang:
so that we get a consistent sql timeout error

http-query-timeout-delta = "2s"


chitang:
I think I should remove max-db-query-timeout in QC.

catstavi:
so to summarize what I see in this PR:
this introduces a small delta that we subtract from the default httpTimeout to get the timeout that will apply to the running query. The goal of this delta is to make it more likely to get a consistent sql timeout error instead of the http connection being closed/timing out.

is that right?

chitang:
Yes

catstavi (approved):

PR #469, opened 2022-04-29
EN-52975: proactively drop unwanted indexes when !secondary-reindex

chitang:


chitang:
Previously, it skips those we don't want to create.  But if they already exist and we are not starting from fresh, these unwanted indexes remain with the dataset.

MarkWillisford (approved):

MarkWillisford (approved):

PR #475, opened 2022-05-19
EN-52985: Switch to use savepoint to reset SQL connection when create…

chitang:
… index encounters error

clearWarnings seems to work with auto commit only

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/store/SqlIndexCleanup.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/store/SqlIndexCleanup.scala
@@ -2,6 +2,7 @@ package com.socrata.pg.store
 
 import java.sql.SQLException
 import com.rojoma.simplearm.v2._
+import com.socrata.pg.query.QueryResult.QueryRuntimeError
 
 import javax.sql.DataSource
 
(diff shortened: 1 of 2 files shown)


common-pg/src/main/scala/com/socrata/pg/store/SqlIndexCleanup.scala
```diff
@@ -37,8 +38,13 @@ class SqlIndexCleanup() extends IndexCleanup {
...
               case ex: SQLException =>
-                conn.clearWarnings()
-                log.error(s"sql error dropping index $name ${ex.toString}")
+                QueryRuntimeError(ex) match {
+                  case None =>
+                    log.error(s"SQL Exception drop index $name (${ex.getSQLState})", ex)
+                  case Some(e) =>
+                    log.warn(s"SQL Exception drop index $name ${e.getClass.getSimpleName} (${ex.getSQLState})")
+                }
+                conn.clearWarnings() // reset connection on the basis of autoCommit being true
```
sjzavaleta, line 47: You say the clearWarnings only seems to work with autoCommit. I see that it has been left here but removed from indexManager. Is that because one path is associated with autoCommit?

chitang: Yes, that's the reason we can keep using it in index drop.

sjzavaleta (approved):

PR #477, opened 2022-05-20
EN-53459, EN-53465: Drop index directives indirectly (pending queue) to avoid record lock problems  index_directives were lost in pg-secondary after publication cycle.

chitang:


chitang:
This is about index_directives which control standard indexes (vs custom indexes).  We learn something about locks in custom indexes and retrofit this in index_directives.

common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
```diff
@@ -140,18 +140,22 @@ class SecondarySchemaLoader[CT, CV](conn: Connection, dsLogger: Logger[CT, CV],
...
 
   protected def dropIndexes(columnInfos: Iterable[ColumnInfo[CT]]): Unit = {
     val table = tableName(columnInfos)
-    using(conn.createStatement()) { stmt =>
+    using(conn.prepareStatement(pendingIndexDropSql)) { stmt =>
       for {
         ci <- columnInfos
-        indexSql <- repFor(ci).dropIndex(table)
-      } stmt.execute(indexSql)
+        dbIndexName <- repFor(ci).dbIndexNames(table)
```
sjzavaleta, line 154: dbIndexNames is plural, might it return more than one val?

common-pg/src/main/scala/com/socrata/pg/store/index/Indexable.scala
```diff
@@ -74,6 +76,17 @@ trait TextIndexable[T] extends Indexable[T] { this: SqlColumnCommonRep[T] =>
...
   }
+
+  override def dbIndexNames(tableName: String): Seq[String] = {
+    this.physColumns.flatMap { phyCol =>
+      Seq(s"idx_${tableName}_$phyCol",
+          s"idx_${tableName}_u_$phyCol",
+          s"idx_${tableName}_nl_$phyCol",
+          s"idx_${tableName}_nf_$phyCol",
+          s"idx_${tableName}_unl_$phyCol",
+          s"idx_${tableName}_unf_$phyCol")
```
sjzavaleta, line 87: For my benefit, what are _u_, _nl_, _nf_, _unl_, _unf_?

common-pg/src/main/scala/com/socrata/pg/store/SecondarySchemaLoader.scala
chitang: Yes, a column has can have more than one standard indexes.

sjzavaleta (approved):

PR #485, opened 2022-06-15
EN-53459: Revert async standard index drop

chitang:


catstavi:
Chi, can we add some info to this that will help us if we forget about all this work and try to come back to it in a year?

I can start. I'll add the info:
We are reverting this due to it causing an issue with the external API for creating indexes. That API (used by DSMAPI in dataset update workflows) expects to be able to drop and index, and then immediately recreate it. Instead it was dropping it, which triggered the drop to be scheduled, and then trying to recreate it (and getting an error instead).

Could you add the info for:
what was the purpose of EN-53459 in the first place/problem it was trying to solve or improve?

catstavi (approved):

PR #496, opened 2022-07-27
EN-54993: SQL generation fails causing rollup not being applied when …

chitang:
…rollup has a column type (like boolean) where the type does not exist in the dataset nor appear in select


chitang:
Previously, we only create for type reps that are in the query SELECT.  If you have
SELECT number, text, there will be no boolean type rep which is normally okay until rollup query rewrite comes in and possibly changes the query to "SELECT .... WHERE boolean_column".  Oops, we didn't expect boolean.  Now we always build typeReps for every datatype.

MarkWillisford (approved):

PR #498, opened 2022-08-08
EN-55111: Rollup application at joins

chitang:


chitang:
Previously
1. rollups are only applied at the prefix of a query tree.
2. only one rollup can be applied in a query.
New
1. rollups can be applied at joins.
2. more than one rollups can be applied in a query.

common-pg/src/main/scala/com/socrata/pg/query/QueryServerHelper.scala
```diff
@@ -105,23 +107,44 @@ object QueryServerHelper {
...
+      (Map[TableName, CopyInfo], Map[TableName, LocalRollupInfo]) = {
+    val rollupRx = "(_.{4}-.{4})\\.(.+)".r
+    tableNames.foldLeft((Map.empty[TableName, CopyInfo], Map.empty[TableName, LocalRollupInfo])) { (acc, tableName) =>
+      tableName.name match {
+        case rollupRx(resourceName, rollupName) =>
+          val ruOpt = for {
+            datasetInfo <- pgu.datasetMapReader.datasetInfoByResourceName(ResourceName(resourceName))
+            copyInfo <- pgu.datasetMapReader.lookup(datasetInfo, LifecycleStage.Published)
+            rollupInfo <- pgu.datasetMapReader.rollup(copyInfo, new RollupName(rollupName))
+          } yield { rollupInfo }
```
sjzavaleta, line 132: Side comment: When do you decide to use for/yield instead of map()?

soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
```diff
@@ -381,11 +381,20 @@ class QueryServer(val dsInfo: DSInfo, val caseSensitivity: CaseSensitivity, val
     }
   }
 
+  private def writeRollupHeader(header: String, ru: Option[RollupName], rus: Seq[RollupInfo])(resp: HttpServletResponse): Unit = {
+    val combined = ru.toSeq.map(_.underlying) ++
+      rus.map(ru => s"${(new TableName(ru.copyInfo.datasetInfo.resourceName.get)).nameWithoutPrefix}.${ru.name.underlying}")
```
sjzavaleta, line 386: can resourceName be empty?

common-pg/src/main/scala/com/socrata/pg/query/QueryServerHelper.scala
chitang: A for comprehension is multiple flatmaps and map.

soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
chitang: resourceName was added after the initial nbe architecture.  For ease of transition, it wasn't made required in the database scheme.  But it is essentially required.

common-pg/src/main/scala/com/socrata/pg/query/QueryServerHelper.scala
```diff
@@ -105,23 +107,44 @@ object QueryServerHelper {
...
-    allMapNoAliases
+  def removeTableAlias(tableNames: Seq[TableName]): Seq[TableName] = {
+    tableNames.map(_.copy(alias = None))
+  }
+
+  def getCopyAndRollupMaps(pgu: PGSecondaryUniverse[SoQLType, SoQLValue],
+                           tableNames: Seq[TableName],
+                           reqCopy: Option[String]):
+      (Map[TableName, CopyInfo], Map[TableName, LocalRollupInfo]) = {
+    val rollupRx = "(_.{4}-.{4})\\.(.+)".r
```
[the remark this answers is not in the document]
chitang: I moved the regex outside the function to the object.  An example is added (repeated from somewhere else).  That should make it clearer.  It is still pretty simple at this stage and hopefully stay this way.

sjzavaleta (approved):

PR #500, opened 2022-09-14
Paul/en 54245

rjmac:


common-pg/src/main/scala/com/socrata/pg/store/LocalRollupInfo.scala
```diff
@@ -5,8 +5,8 @@ import java.security.MessageDigest
 import com.socrata.datacoordinator.id.RollupName
 import com.socrata.datacoordinator.truth.metadata.{CopyInfo, RollupInfo}
 
-class LocalRollupInfo(copyInfo: CopyInfo, name: RollupName, soql: String, val tableName: String)(implicit tag: com.socrata.datacoordinator.truth.metadata.`-impl`.Tag) extends RollupInfo(copyInfo, name, soql, None) {
-  def updateName(newTableName: String) = new LocalRollupInfo(copyInfo, name, soql, newTableName)
+class LocalRollupInfo(copyInfo: CopyInfo, name: RollupName, soql: String, val tableName: String,val systemId: Long = -1L)(implicit tag: com.socrata.datacoordinator.truth.metadata.`-impl`.Tag) extends RollupInfo(copyInfo, name, soql, None) {
```
rjmac, line 8: I generally prefer to have an `Option[Long]` over a magic value, though I'm not sure this even needs to be optional in any sense... (see further comments below)

Also I prefer the various kinds of Ids to have types rather than just being bare numbers.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -2,52 +2,155 @@ package com.socrata.pg.store
...
   def localRollupsQuery = "SELECT name, soql, table_name FROM rollup_map WHERE copy_system_id = ?"
+
   override def rollups(copyInfo: CopyInfo): Seq[LocalRollupInfo] = {
     using(conn.prepareStatement(localRollupsQuery)) { stmt =>
       stmt.setLong(1, copyInfo.systemId.underlying)
       using(t("rollups", "copy_id" -> copyInfo.systemId)(stmt.executeQuery())) { rs =>
         val res = Vector.newBuilder[LocalRollupInfo]
-        while(rs.next()) {
+        while (rs.next()) {
           res += new LocalRollupInfo(copyInfo, new RollupName(rs.getString("name")), rs.getString("soql"), rs.getString("table_name"))
```
rjmac, line 24: This should probably pull out the system id

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -2,52 +2,155 @@ package com.socrata.pg.store
...
+              rs.getLong("copy_map_copy_number"),
+              LifecycleStage.valueOf(rs.getString("copy_map_lifecycle_stage")),
+              rs.getLong("copy_map_data_version"),
+              rs.getLong("copy_map_data_shape_version"),
+              new DateTime(rs.getTimestamp("copy_map_last_modified").getTime),
+              Some(rs.getLong("copy_map_table_modifiers_table_modifier"))
+            )
+          })
+          .toSet
+      }
```
rjmac, line 88: I'm not a huge fan of this kind of interleaved side-effects on iterators.  I prefer to just embrace the imperativity and use a while loop.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
```diff
@@ -148,29 +148,72 @@ class PGSecondaryDatasetMapWriter[CT](override val conn: Connection,
...
       case Some(_) =>
         using(conn.prepareStatement(updateLocalRollupQuery)) { stmt =>
           stmt.setString(1, soql)
           stmt.setString(2, tableName)
           stmt.setString(3, name.underlying)
           stmt.setLong(4, copyInfo.systemId.underlying)
-          t("create-or-update-rollup", "action" -> "update", "copy-id" -> copyInfo.systemId, "name" -> name)(stmt.executeUpdate())
+          t("create-or-update-rollup", "action" -> "update", "copy-id" -> copyInfo.systemId, "name" -> name)(
+            using(stmt.executeQuery()){resultSet =>
+              if(resultSet.next()){
```
rjmac, line 165: This (and the None branch) should be much more unhappy if it somehow fails to have a row returned.  That'd also let the `match` just return the systemId instead of setting it as a side-effect.

rjmac (commented):
I generally like the look of this.  My only high-level concerns are:
* Making sure that a "bad" rollup doesn't break replication
* Making sure that if two datasets are updated simultaneously then the rollups are kept up to date correctly.  Think this order of events, and imagine that both of these are :
  1. Replication of dataset A (which has a preexisting join-rollup to dataset B) begins
  2. Replication of dataset B (which has a preexisting join-rollup to dataset A) begins
  3. A updates its rollups
  4. B updates its rollups
  5. A commits
  6. B commits
  so neither dataset's updates are visible to the other while the replication is in progress.
* Making sure everything works out when multiple copies are involved.  This is actually an annoying case that I just thought about because
  1. Working copies should _not_ have populated rollups, except...
  2. After a SecondaryReindex event has been received (this rebuilds indexes and rollups before the publication event that will be coming shortly arrives so the publication replication is fast).

rjmac:
Closing this PR so I can approve the real one :)

PR #502, opened 2022-09-20
EN-55557: Add un-casefolded tpo indexes to text columns

rjmac:


common-pg/src/main/scala/com/socrata/pg/store/index/Indexable.scala
```diff
@@ -31,6 +31,10 @@ trait TextIndexable[T] extends Indexable[T] { this: SqlColumnCommonRep[T] =>
          |    CREATE INDEX idx_${tableName}_u_${phyCol} on ${tableName}
          |    USING BTREE (upper(${phyCol}) text_pattern_ops)${tablespace};
          |  END IF;
+         |  IF NOT EXISTS(select 1 from pg_indexes WHERE indexname = 'idx_${tableName}_tpo_${phyCol}') THEN
+         |    CREATE INDEX idx_${tableName}_nl_${phyCol} on ${tableName}
```
sjzavaleta, line 35: _nl_ -> _tpo_ ?

rjmac: Oops yep!

sjzavaleta (approved):

PR #505, opened 2022-09-23
EN-54245 Fix for RowUpdatedPreview

PaulMTT:
On the handling metadata delete scenario, turns out I was deleting in a bad code place.
I did not account for a RowUpdatePreview event, which was causing the rollup metadata to be deleted when a rollup was updated (new metadata in rollup_relationship_map).

I removed the deletion from that code place and instead put it closer to where we delete copies.


PaulMTT:
Guess me repushing/creating this branch with already merged commits, then fixing the conflict caused the history to undelete that one line? Looks good now.

branweb1 (approved):

PR #507, opened 2022-10-25
Fix failure to populate rollup in certain cases

rjmac:


common-pg/src/main/scala/com/socrata/pg/query/QueryServerHelper.scala
```diff
@@ -159,7 +159,7 @@ object QueryServerHelper {
   }
 
   def getCopy(pgu: PGSecondaryUniverse[SoQLType, SoQLValue], datasetInfo: DatasetInfo, reqCopy: Option[String]): CopyInfo = {
-    val intRx = "^[0-9]+$".r
+    val intRx = "^([0-9]+)$".r
```
saraid, line 162: That's _insidious_.

saraid (approved):
My understanding of why this fixes the issue is a bit shaky, but it still looks reasonable to me to be populating stuff that wasn't getting populated and probably should be.

PR #514, opened 2022-11-16
Feature/joined rollups

PaulMTT:


.run/secondary-watcher-pg.run.xml
```diff
@@ -0,0 +1,17 @@
+<component name="ProjectRunConfigurationManager">
```
PaulMTT, line 1: These are just stored IntelliJ run configs, to make things easier for others in the future.

common-pg/src/main/scala/com/socrata/Predicates.scala
```diff
@@ -0,0 +1,29 @@
...
+
+    def &&(them: Predicate[T]): Predicate[T] = (t: T) => me(t) && them(t)
+
+    def |(them: Predicate[T]): Predicate[T] = (t: T) => me(t) | them(t)
+
+    def ||(them: Predicate[T]): Predicate[T] = (t: T) => me(t) || them(t)
+
+    // this allows you to specify negation of a predicate before its applied
+    // e.g (!isAlphaNumeric && !isUnderscore)('a')
+    def unary_!(c: T): Boolean = !me(c)
```
zachlefevre, line 19: This fn cannot be invoked.
You're demanding a T in order to invoke this ! fn. That means you'll need to provide an argument to invoke !.
You won't be able to write !foo && !bar.

I'd recommend using this pattern https://coderwall.com/p/k_1jzw/scala-s-pimp-my-library-pattern-example
which in your case would be
```
package com.socrata

import scala.language.implicitConversions

object Predicates {

  case class Predicate[T](fn: Function[T, Boolean]) { self =>
    def apply(t: T) = fn(t)
    def ||(them: Predicate[T]): Predicate[T] = Predicate(t => self(t) || them(t))
    def &&(them: Predicate[T]): Predicate[T] = Predicate(t => self(t) && them(t))
    def unary_!(): Predicate[T] = Predicate(t => !self(t))
  }

  implicit def fnTopredicate[T](fn: T => Boolean): Predicate[T] = Predicate(fn)
  implicit def predicateToFn[T](pred: Predicate[T]) = pred.fn

  def isAlphaNumericUnderscore: Predicate[Char] = isAlphaNumeric || isUnderscore

  def isAlphaNumeric: Predicate[Char] = Character.isLetterOrDigit _

  def isUnderscore: Predicate[Char] = (_: Char) == '_'

}
```

common-pg/src/main/scala/com/socrata/pg/store/LocalRollupInfo.scala
```diff
@@ -1,22 +1,21 @@
...
-    val nameHash = sha1.digest(name.underlying.getBytes("UTF-8")).take(bytes).map("%02x" format _).mkString
-
-    copyInfo.dataTableName + "_r_" + copyInfo.dataVersion + "_" + nameHash
+  // Generate the name for the actual rollup table
+  // It is important that this is somewhat random so that we can drop/recreate a rollup table and not have names conflict
+  // Note that table name should be lowercase, as this is an assumption throughout the stack
+  def tableName(copyInfo: CopyInfo, name: RollupName, sequenceToAppend: String): String = {
+    val base = copyInfo.dataTableName + "_r_" + copyInfo.dataVersion + "_" + name.underlying.filter(Predicates.isAlphaNumericUnderscore).toLowerCase + "_"
+    // clamp our base string to be max 63-randomBits since postgres table names are limited to 63 characters
+    (base.take(63 - sequenceToAppend.length) + sequenceToAppend)
```
zachlefevre, line 19: unnecessary parentheses

common-pg/src/main/scala/com/socrata/Predicates.scala
PaulMTT: Oh you are right that `!something && !other` fails.
The solution you pasted works. 
Ill update this.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,167 @@
...
 
-trait LocalRollupReaderOverride[CT] { self: BasePostgresDatasetMapReader[CT] =>
-  def localRollupsQuery = "SELECT name, soql, table_name FROM rollup_map WHERE copy_system_id = ?"
+trait LocalRollupReaderOverride[CT] {
+  self: BasePostgresDatasetMapReader[CT] =>
   override def rollups(copyInfo: CopyInfo): Seq[LocalRollupInfo] = {
     using(conn.prepareStatement(localRollupsQuery)) { stmt =>
       stmt.setLong(1, copyInfo.systemId.underlying)
       using(t("rollups", "copy_id" -> copyInfo.systemId)(stmt.executeQuery())) { rs =>
         val res = Vector.newBuilder[LocalRollupInfo]
```
zachlefevre, line 19: All of this mutable code can be made much nicer with something like 
```
  private def asIteratorOf[T](stmt: PreparedStatement)(f: ResultSet => T): Iterator[T] = {
    val resultSet = stmt.executeQuery()

    new Iterator[T] {
      def hasNext = resultSet.next()
      def next() = f(resultSet)
    }
  }

  def asVectorOf[T](stmt: PreparedStatement)(f: ResultSet => T): Vector[T] = asIteratorOf(stmt)(f).toVector
  ```

File: common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala [hunk 5]
```diff
@@ -1,53 +1,167 @@
...
     }
   }
 
-  def localRollupQuery = "SELECT soql, table_name FROM rollup_map WHERE copy_system_id = ? AND name = ?"
+  def localRollupsQuery = "SELECT system_id, name, soql, table_name FROM rollup_map WHERE copy_system_id = ?"
+
   override def rollup(copyInfo: CopyInfo, name: RollupName): Option[LocalRollupInfo] = {
     using(conn.prepareStatement(localRollupQuery)) { stmt =>
       stmt.setLong(1, copyInfo.systemId.underlying)
       stmt.setString(2, name.underlying)
```
zachlefevre, line 33: your query is on longer dependent on a second param, but you're passing in a second param.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,167 @@
...
-      using(t("rollup", "copy_id" -> copyInfo.systemId,"name" -> name)(stmt.executeQuery())) { rs =>
-        if(rs.next()) {
-          Some(new LocalRollupInfo(copyInfo, name, rs.getString("soql"), rs.getString("table_name")))
+      using(t("rollup", "copy_id" -> copyInfo.systemId, "name" -> name)(stmt.executeQuery())) { rs =>
+        if (rs.next()) {
+          Some(new LocalRollupInfo(copyInfo, name, rs.getString("soql"), rs.getString("table_name"), RollupId(rs.getLong("system_id"))))
```
zachlefevre, line 36: why are we not using the name from the DB?

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,167 @@
...
         } else {
           None
         }
       }
     }
   }
+
+  def localRollupQuery = "SELECT system_id, soql, table_name FROM rollup_map WHERE copy_system_id = ? AND name = ?"
```
zachlefevre, line 44: can we move this up to where localRollupsQuery (note the plural) so it's obvious this query is being executed and not the localRollupsQuery?

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,167 @@
...
+
+import java.sql.Connection
+import scala.reflect.ClassTag
 
-trait LocalRollupReaderOverride[CT] { self: BasePostgresDatasetMapReader[CT] =>
-  def localRollupsQuery = "SELECT name, soql, table_name FROM rollup_map WHERE copy_system_id = ?"
+trait LocalRollupReaderOverride[CT] {
+  self: BasePostgresDatasetMapReader[CT] =>
   override def rollups(copyInfo: CopyInfo): Seq[LocalRollupInfo] = {
     using(conn.prepareStatement(localRollupsQuery)) { stmt =>
```
zachlefevre, line 16: can we move the localRollupsQuery back up here since this is where it's being executed?

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,167 @@
...
           None
         }
       }
     }
   }
+
+  def localRollupQuery = "SELECT system_id, soql, table_name FROM rollup_map WHERE copy_system_id = ? AND name = ?"
+
+  def getRollupCopiesRelatedToCopy(copyInfo: CopyInfo): Set[CopyInfo] = {
+    var out = Set[CopyInfo]()
```
zachlefevre, line 47: this mutability can be cleaned up with the asIterator and asVector fns given above. You could write an asSet.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,167 @@
...
+            new DateTime(rs.getTimestamp("copy_map_last_modified").getTime),
+            Some(rs.getLong("copy_map_table_modifiers_table_modifier"))
+          )
+        }
+      }
+    }
+    out
+  }
+
+  def getRollupCopiesRelatedToCopyQuery =
```
zachlefevre, line 68: can we move this to above where it's used?

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,167 @@
...
+      stmt.setString(1, sequenceName)
+      using(t("getNextSequence", "sequence" -> sequenceName)(stmt.executeQuery())) { rs =>
+        if (rs.next()) Some(rs.getObject(1)).asInstanceOf[Option[T]] else None
+      }
+    }
+  }
+
+  def getNextRollupTableNameSequence = getNextSequence[Long](RollupManager.tableNameSequenceIdentifier).getOrElse(throw new IllegalStateException(s"Could not get next value of sequence '${RollupManager.tableNameSequenceIdentifier}'"))
+
+  def getRollupsRelatedToCopyQuery =
```
zachlefevre, line 128: this too. Can we move it up?

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
PaulMTT: Not my method, and did not change this much.
I have previously pitched using a syntax like:
`Iterator.continually(rs.next())
  .takeWhile(identity)
  .map(_ => Model(
      id = rs.getInt("id"),
      text = rs.getString("text")
   ))`
   
   But this was turned down in other PRs in favor of simplicity.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,167 @@
...
+    }
+    out
+  }
+
+  def nextSequenceValue = "select nextval(?)"
+  def getNextSequence[T](sequenceName: String): Option[T] = {
+    using(conn.prepareStatement(nextSequenceValue)) { stmt =>
+      stmt.setString(1, sequenceName)
+      using(t("getNextSequence", "sequence" -> sequenceName)(stmt.executeQuery())) { rs =>
+        if (rs.next()) Some(rs.getObject(1)).asInstanceOf[Option[T]] else None
```
zachlefevre, line 121: `        if (rs.next()) Some(rs.getObject(1).asInstanceOf[T]) else None` is more direct. it's the object we're casting, not the Option.

PaulMTT: Has to be optional because you cannot guarantee a primitive will exist, or that the sequence exists.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
```diff
@@ -148,29 +150,71 @@ class PGSecondaryDatasetMapWriter[CT](override val conn: Connection,
...
+                return Some(new LocalRollupInfo(copyInfo, name, soql, tableName,RollupId(resultSet.getLong("system_id"))))
+              }
+            }
+          )
         }
     }
-    new LocalRollupInfo(copyInfo, name, soql, tableName)
+    None
+  }
+  def createRollupRelationship = "insert into rollup_relationship_map (rollup_system_id, referenced_copy_system_id) values (?,?) on conflict do nothing"
```
zachlefevre, line 189: does not need to be a def

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
```diff
@@ -148,29 +150,71 @@ class PGSecondaryDatasetMapWriter[CT](override val conn: Connection,
...
+    using(conn.prepareStatement(createRollupRelationship)) { stmt =>
+      stmt.setLong(1, rollupInfo.systemId.underlying)
+      stmt.setLong(2, relatedCopyInfo.systemId.underlying)
+      t("create-rollup-relationship", "rollupInfo" -> rollupInfo.systemId, "referencedCopyId" -> relatedCopyInfo.systemId.underlying)(
+        stmt.executeUpdate()
+      )
+    }
+  }
+
+  def deleteRollupRelationshipsRollupQuery = "delete from rollup_relationship_map where rollup_system_id = ?"
```
zachlefevre, line 200: same here

PaulMTT:
Okay I will unformed my auto format to keep query defs close to usage.

common-pg/src/main/scala/com/socrata/Predicates.scala
PaulMTT: Is it possible to do this without a wrapping object on the function?

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
PaulMTT: def is how we declare the other query strings.
I agree this could be a val.
I guess the benefit of a def is that its not stored in memory?
Probably not really a concern either way.
I'd prefer to keep to similar to the rest of the code style.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
```diff
@@ -148,29 +150,71 @@ class PGSecondaryDatasetMapWriter[CT](override val conn: Connection,
...
-          stmt.setString(2, tableName)
-          stmt.setString(3, name.underlying)
-          stmt.setLong(4, copyInfo.systemId.underlying)
-          t("create-or-update-rollup", "action" -> "update", "copy-id" -> copyInfo.systemId, "name" -> name)(stmt.executeUpdate())
+          stmt.setString(2, name.underlying)
+          stmt.setLong(3, copyInfo.systemId.underlying)
+          t("create-or-update-rollup", "action" -> "update", "copy-id" -> copyInfo.systemId, "name" -> name)(
+            using(stmt.executeQuery()){resultSet =>
+              if(resultSet.next()){
+                return Some(new LocalRollupInfo(copyInfo, name, soql, resultSet.getString("table_name"),RollupId(resultSet.getLong("system_id"))))
```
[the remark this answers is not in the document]
zachlefevre: and it'd be even better if you had a function like `def asOptionOf[T](stmt: PreparedStatement)(f: ResultSet => T): Vector[T] = ???` so you can get rid of the ifs entirely.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
```diff
@@ -148,29 +150,71 @@ class PGSecondaryDatasetMapWriter[CT](override val conn: Connection,
...
         using(conn.prepareStatement(insertLocalRollupQuery)) { stmt =>
           stmt.setString(1, name.underlying)
           stmt.setLong(2, copyInfo.systemId.underlying)
           stmt.setString(3, soql)
           stmt.setString(4, tableName)
-          t("create-or-update-rollup", "action" -> "insert", "copy-id" -> copyInfo.systemId, "name" -> name)(stmt.executeUpdate())
+          t("create-or-update-rollup", "action" -> "insert", "copy-id" -> copyInfo.systemId, "name" -> name)(
+            using(stmt.executeQuery()){resultSet =>
+              if(resultSet.next()){
+                return Some(new LocalRollupInfo(copyInfo, name, soql, tableName,RollupId(resultSet.getLong("system_id"))))
```
zachlefevre, line 181: we can get rid of the returns here as well.

common-pg/src/main/scala/com/socrata/pg/store/RollupId.scala
```diff
@@ -0,0 +1,21 @@
...
+  implicit val jCodec = new JsonDecode[RollupId] with JsonEncode[RollupId] {
+    def encode(rollupId: RollupId) = JNumber(rollupId.underlying)
+
+    def decode(v: JValue): Either[DecodeError, RollupId] = v match {
+      case n: JNumber => Right(new RollupId(n.toLong))
+      case other => Left(DecodeError.InvalidType(JNumber, other.jsonType))
+    }
+  }
+
+  val Invalid = new RollupId(-1)
```
zachlefevre, line 20: I would get rid of this until it's used.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
zachlefevre: the name is in the table, but we ignore the name from the table and instead use the name from local scope. Do you know why that is?

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
zachlefevre: I think you can. You'll need to more the asInstanceOf into the Some.
Try compiling with the exact line I gave above.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
zachlefevre: fair enough. A def allows for parameters, which you aren't using, nor are you relying on the laziness of an unevaluated fn.
up2u :)

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
PaulMTT: Oh I see what you mean!
Yeah it makes sense to change that.

store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -522,9 +537,22 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
       }
     }
 
+    def updateRelatedRollups(copyInfo: TruthCopyInfo): Unit ={
+      pgu.datasetMapReader.getRollupCopiesRelatedToCopy(copyInfo).foreach(relatedCopy=>{
+        logger.info(s"Updating rollups for ${relatedCopy.systemId}, as they relate to current ${copyInfo.systemId}")
+        updateRollups(pgu,
+          Some(relatedCopy),
+          relatedCopy,
+          tryToMove = { _ =>false},
```
zachlefevre, line 550: unnecessary brackets

common-pg/src/main/scala/com/socrata/Predicates.scala
zachlefevre: it'll only work for predicates, not all functions, and only where this is imported.

zachlefevre: I think the PML pattern can get pretty scary pretty quickly. There's lots of hidden conversions that can occur and chew through your performance.
I'm not sure of another way around this.

zachlefevre: btw scala3 has a much nicer mechanism for this.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
zachlefevre: that wouldn't quite work as takeWhile is assuming you're only emitting booleans.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
PaulMTT: Same as other comment.
I will say that I prefer using the expressiveness of Scala over simplicity in these scenarios.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapWriter.scala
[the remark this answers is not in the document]
PaulMTT: Fair enough, I'm in favor of removing returns and shifting the none

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
PaulMTT: resultset.next is a boolean, and identity function just echoes that back as the continue condition

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
PaulMTT: oh wait I misread this. Okay I agree

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
PaulMTT: Okay I'm with you on making this change to limit mutability.

common-pg/src/main/scala/com/socrata/Predicates.scala
PaulMTT: Okay ill need a minute on this one, I'm not happy with wrapping the function in an object

PaulMTT: Interesting...it can work, BUT
It comes down to `unary_!` needs to be a method(val) instead of a function(def).
Further complicated by scala casting between function and methods, and generics.

PaulMTT: If we leave generics in place, but turn unary_! into a method, then ->
```
value || is not a member of Char => Boolean
(!isAlphaNumeric || isUnderscore)('!') 
```

PaulMTT: If you turn `unary_!` into a method, but replace generic Predicate[T] and (Char)=>Boolean, it works out.

PaulMTT: Ah the key is that the `me` of the implicit needs to be completely unwrapped and raw
```
implicit class PredicateOperators[T](me: (T)=>Boolean) 
```

PaulMTT:
```
type Predicate[T] = (T)=>Boolean

implicit class PredicateOperators[T](me: (T)=>Boolean) {

  def &(them: Predicate[T] ): Predicate[T]  = (t: T) => me(t) & them(t)

  def &&(them: Predicate[T] ): Predicate[T]  = (t: T) => me(t) && them(t)

  def |(them: Predicate[T] ): Predicate[T]  = (t: T) => me(t) | them(t)

  def ||(them: Predicate[T] ): Predicate[T]  = (t: T) => me(t) || them(t)

  // this allows you to specify negation of a predicate before its applied
  // e.g (!isAlphaNumeric && !isUnderscore)('a')
  val unary_! = (c:T) => !me(c)

}

def isAlphaNumeric: Predicate[Char] = Character.isLetterOrDigit

def isUnderscore: Predicate[Char] = _ == '_'

def isAlphaNumericUnderscore: Predicate[Char] = isAlphaNumeric || isUnderscore

(isAlphaNumeric || isUnderscore)('a')
(isAlphaNumeric || isUnderscore)('_')
(isAlphaNumeric || isUnderscore)('!')
(isAlphaNumeric || !isUnderscore)('!')
(!isAlphaNumeric || isUnderscore)('!')
(!isAlphaNumeric || !isUnderscore)('a')

```

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,174 @@
...
+              rs.getString("rollup_map_soql"),
+              rs.getString("rollup_map_table_name"),
+              RollupId(rs.getLong("rollup_map_system_id")))
+          ).toSet
+      }
+    }
+  }
+
+  def nextSequenceValue = "select nextval(?)"
+  def getNextSequence[T](sequenceName: String): Option[T] = {
```
rjmac, line 149: This type is a lie; it says "I can give you any type T that you ask for".

But sequences always return that are smaller than `bigint`, so this T must (a) always be such an integer, and (b) can always losslessly be `Long`.

rjmac: s/smaller than/no larger than/

common-pg/src/main/resources/com/socrata/pg/store/schema/20221116-create-rollup-table-name-seq.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <changeSet author="Paul Maksimovich" id="20221116-create-rollup-table-name-seq">
+        <sql>
+            create sequence rollup_table_name_seq
+                minvalue 0
+                cycle;
```
rjmac, line 10: It will be quite the accomplishment if we ever flow a bigint here!

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
PaulMTT: Oh! True, let me fix that up.
But correct my intent was to allow any of the sequence types.

common-pg/src/main/resources/com/socrata/pg/store/schema/20221116-create-rollup-table-name-seq.xml
PaulMTT: But if and when it does I want it!

PaulMTT:
Ooop borked something during a refactor, failing tests. Once sec.

rlvoyer (approved):
🥇 Looks good to me.

rlvoyer:
Pls don't forget to squash!

store-pg/src/test/scala/com/socrata/pg/store/DatabaseTestBase.scala
```diff
@@ -223,7 +222,7 @@ object DatabaseTestBase {
 
   private def createDatabase(dbName: String): Unit = {
     try {
-      Class.forName("org.postgresql.Driver").newInstance()
+      Class.forName("org.postgresql.Driver").getDeclaredConstructor().newInstance()
```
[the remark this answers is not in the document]
PaulMTT: This is deprecated in the future (java 9). I found out when running with a newer runtime, and just kept the change.

common-pg/src/main/scala/com/socrata/pg/store/PGSecondaryDatasetMapReader.scala
```diff
@@ -1,53 +1,174 @@
...
+
+  def getRollupCopiesRelatedToCopy(copyInfo: CopyInfo): Set[CopyInfo] = {
+    using(conn.prepareStatement(getRollupCopiesRelatedToCopyQuery)) { stmt =>
+      stmt.setLong(1, copyInfo.systemId.underlying)
+      using(t("getRollupCopiesRelatedToCopy", "copy_id" -> copyInfo.systemId)(stmt.executeQuery())) { rs =>
+        Iterator.continually(rs.next())
+          .takeWhile(identity)
+          .map(_=>
+            CopyInfo(
+              DatasetInfo(new DatasetId(rs.getLong("dataset_map_system_id")), rs.getLong("dataset_map_next_counter_value"), rs.getLong("dataset_map_latest_data_version"), rs.getString("dataset_map_locale_name"), rs.getBytes("dataset_map_obfuscation_key"), Option(rs.getString("dataset_map_resource_name"))),
```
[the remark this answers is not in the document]
PaulMTT: I completely agree! This was one of my gripes with the existing way we do our mappings. Definitely should be a future work to refactor our mapping code into reusable helpers.

common-pg/src/main/scala/com/socrata/pg/store/LocalRollupInfo.scala
```diff
@@ -1,22 +1,21 @@
...
-import java.security.MessageDigest
-
+import com.socrata.Predicates
 import com.socrata.datacoordinator.id.RollupName
 import com.socrata.datacoordinator.truth.metadata.{CopyInfo, RollupInfo}
 
-class LocalRollupInfo(copyInfo: CopyInfo, name: RollupName, soql: String, val tableName: String)(implicit tag: com.socrata.datacoordinator.truth.metadata.`-impl`.Tag) extends RollupInfo(copyInfo, name, soql, None) {
-  def updateName(newTableName: String) = new LocalRollupInfo(copyInfo, name, soql, newTableName)
+
+class LocalRollupInfo(copyInfo: CopyInfo, name: RollupName, soql: String, val tableName: String, val systemId: RollupId)(implicit tag: com.socrata.datacoordinator.truth.metadata.`-impl`.Tag) extends RollupInfo(copyInfo, name, soql, None) {
```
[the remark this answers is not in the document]
PaulMTT: Honestly I do not fully understand that syntax. I believe the tags are some sort of assertion?
I'll read more about this functionality.

PR #518, opened 2022-12-07
EN-55757 rollup usage metrics

PaulMTT:
This change brings in a new library(rollup-metrics) and 
* logs the creation time and size of a rollup table that was built.
* logs when a rollup is hit

### DIFF
--- a/common-pg/build.sbt
+++ b/common-pg/build.sbt
@@ -26,7 +26,8 @@ libraryDependencies ++= Seq(
   metricsJetty,
   metricsGraphite,
   metricsJmx,
-  scalatest % "test"
+  scalatest % "test",
+  rollupMetrics
 )
 
 enablePlugins(BuildInfoPlugin)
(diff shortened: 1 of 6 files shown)


rlvoyer:
Ah, looks like you have some failing checks to attend to, @PaulMTT

PaulMTT:
> This looks good to me.
> 
> So the new call to report the pg_total_relation_size happens once upon rollup creation. This seems like a reasonable performance hit. Though if we found that it was problematic to hold up the completion of this `updateRollup` method, we could potentially this off in a Future and log the result as it comes back.
> 
> And presumably, there are no other places outside of `updateRollup` where we need to do this logging?

Yup correct on both. We can keep that in mind, and shift it to a future or something if it's a problem. 
Pretty certain this is the only place where its built. I'll be making a PR to query-coordinator next for the read side.

PaulMTT:
> Ah, looks like you have some failing checks to attend to, @PaulMTT

Turns out upgrading our slf4j is more complicated than I wanted. So I've downgraded the rollup-metrics slf4j to match soql-postgres-adapter. The build errors were from duplicate classes brought by reload4j and log4j.

PaulMTT:
Okay yup added read side here. This goes back to how query-coordinator really should not be the one doing rewriting.

rlvoyer (approved):

PR #531, opened 2023-04-07
Fix (I hope!) a really subtle bug involving joined rollups and replic…

rjmac:
…ations

Ok, so what's happening here was this:
 * We observed rollup replication for a particular dataset+rollup failing with a "no such table" postgresql error.
 * Upon investigating, we found that a joined-to (well, unioned-to) dataset was using copy 3 (a discarded copy!) rather than copy 4 (a published copy) of one of the other tables involved in the query.
 * Following the logic through, the query sqlizer uses the requested copy number of the current primary dataset.
 * For queries, this will effectively always be "published" and so it will always look up published copies of joined-to things, which is correct.
 * For replication it will be the current _copy number_ rather than one of the symbolic names.
 * So, since this was replicating, it was looking up copy 3 of this other dataset because we were replicating copy 3 of the base dataset!! (n.b., if it can't find the requested copy it silently switches to the "latest" copy, which is _also_ surprising and probably wrong but this commit doesn't change that; however, it is probable that this further masked the issue)

So what this does is:

 * Keep passing around the requested copy value down to the relevant finder
 * Because we want to use the same copy _if we're doing a self-join of some kind_
 * But only actually _use_ that requested copy if we are in fact doing a self-join, otherwise ask for the published copy of whatever we're joining to.

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/query/QueryServerHelper.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/query/QueryServerHelper.scala
@@ -62,7 +62,7 @@ object QueryServerHelper {
       val copyInfo = readCtx.copyInfo
 
       val relatedTableNames = removeTableAlias(collectRelatedTableNames(analyses))
-      val (relatedCopyMap, _relatedRollupMap) = getCopyAndRollupMaps(pgu, relatedTableNames, reqCopy)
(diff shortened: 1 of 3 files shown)


saraid (approved):
I see no problems.

PR #532, opened 2023-04-13
reduce_precision soql function

PeterAustinMoore:
- Partner to https://github.com/socrata-platform/soql-reference/pull/306
- Bump Soql stdlib version
- Add ST_ReducePrecision call


rjmac (approved):

rjmac (changes requested):
Comment to satisfy github; actual comment over in slack, but TLDR st_reduceprecision can split single geometries into multigeometries, so this needs to be more complex to make soql's typesystem happy.

rjmac (approved):

PR #533, opened 2023-05-05
Allow other secondary types

zachlefevre:


store-pg/docker/secondary.conf.j2
```diff
@@ -55,7 +55,7 @@ instances {
       {% if ZOOKEEPER_ENSEMBLE is defined -%}
       curator = {
         ensemble = {{ ZOOKEEPER_ENSEMBLE }}
-        namespace = "com.socrata/soda"
+        namespace = "com.socrata/soda" // maybe this
```
rjmac, line 58: Shouldn't be necessary to change this

rjmac: but you might want to make service-advertisement.name configurable

PR #545, opened 2023-06-07
Allow metrics prefix to be configurable

zachlefevre:


configs/application-alpha.conf
```diff
@@ -5,4 +5,4 @@ common-instance = alpha
 com.socrata.coordinator.secondary-watcher = {
   database.database = datacoordinator_alpha
   collocation.group = [alpha, bravo, charlie]
-}
```
zachlefevre, line 8: These are missing newlines my editor just found. May as well correct them now.

soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
```diff
@@ -219,7 +219,7 @@ class QueryServer(val dsInfo: DSInfo, val caseSensitivity: CaseSensitivity, val
   private def withRequestLimit(datasetId: String, resp: HttpServletResponse, f: HttpServletResponse => Unit) = {
     val tooManyAlready = currentRequests.synchronized {
       val currentRequestsForDataset = currentRequests(datasetId)
-      logger.trace(s"Currently processing $currentRequestsForDataset for dataset $datasetId, max allowed is $QueryServer.config.maxConcurrentRequestsPerDataset")
+      logger.trace(s"Currently processing $currentRequestsForDataset for dataset $datasetId, max allowed is ${QueryServer.config.maxConcurrentRequestsPerDataset}")
```
zachlefevre, line 222: This was obviously a mistake, may as well fix it now.

configs/application-alpha.conf
PaulMTT: Yup! Confusing because GitHub shows the curly in red

PaulMTT (approved):

PR #558, opened 2023-06-19
Median and linear regression aggregate functions

rjmac:
* median
* median_disc
* regr_intercept
* regr_slope

median_disc is the most interesting one, because in order to support "median_disc(distinct whatever)" the custom aggregate function needed to be implemented for distinct mode.


saraid (approved):

rjmac:
Force-push to change the commit message from "..needed to be implemented for distinct mode" to "..needed to be implemented for discrete mode"; no other changes.

PR #560, opened 2023-06-20
Implement substring/2 and substring/3

rjmac:


common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizer.scala
```diff
@@ -395,6 +407,8 @@ class SoQLFunctionSqlizer[MT <: MetaTypes with ({ type ColumnType = SoQLType; ty
       LeftPad -> sqlizePad("lpad"),
       RightPad -> sqlizePad("rpad"),
       Chr -> sqlizeChr,
+      Substr2 -> sqlizeSubstr(1),
+      Substr2 -> sqlizeSubstr(2),
```
saraid, line 411:
```suggestion
      Substr3 -> sqlizeSubstr(2),
```
right?

saraid (approved):

PR #570, opened 2023-06-26
Offload a bunch of stuff into stored procedures

rjmac:
Pretty much all of this code is very uninteresting, and as SQL it's easier to read than as Doc-building code.


saraid (approved):
I can see no problems.

PR #573, opened 2023-06-30
Make fake-location human_address serialization work _exactly_ in v1 a…

rjmac:
…nd v2

Specifically, this produces a string containing a JSON blob, and v2 was different in two respects:
 * null subcolumns serialize as empty strings (:crying_cat_face:)
 * minor whitespace differences


saraid (approved):
I hate it.

PR #574, opened 2023-07-01
Process literal strings so the pretty printer doesn't erase newlines

rjmac:


saraid (approved):
No better way to express "don't touch this"?

rjmac:
> No better way to express "don't touch this"?

Nope.  That's just how the pretty printer interprets newlines (and that's because the prettyprinter Haskell library it's a port of does it that way) but in general I think it's not an unreasonable design.

PR #579, opened 2023-07-10
Fix typo'd sql function name

rjmac:


saraid (approved):
Can't believe I missed that. 🤦

PR #585, opened 2023-07-17
Implement interval rep

rjmac:
Also bump the postgresql jdbc driver to the latest so that PGInterval has all the methods I want to call on it.


saraid (approved):
I can't believe I'm approving the version upgrade.

PR #587, opened 2023-07-27
EN-60871: Lock datasets involved in joins against concurrent updates

rjmac:
When performing a replication event, take locks so that concurrent updates on joined datasets always see a consistent view.

### DIFF
--- a/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
+++ b/store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
@@ -1,27 +1,31 @@
 package com.socrata.pg.store
 
+import scala.collection.immutable.SortedSet
+import scala.concurrent.duration._
+
 import java.io.Closeable
 import java.util.concurrent.{CountDownLatch, TimeUnit}
 
 import com.socrata.curator.CuratorFromConfig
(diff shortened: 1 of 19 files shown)


store-pg/src/main/scala/com/socrata/pg/store/PGSecondary.scala
```diff
@@ -303,10 +380,14 @@ class PGSecondary(val config: Config) extends Secondary[SoQLType, SoQLValue] wit
...
-            // Either this copy or some newer copy
-            allCopies.find(existingCopyInfo => existingCopyInfo.copyNumber >= copyInfo.copyNumber)
+          val theCopy = existingDataset match {
+            case Some(dsInfo) =>
+              takeRollupLocks(pgu, dsInfo, upcomingRollups)
+              val allCopies = pgu.datasetMapReader.allCopies(dsInfo)
+              // Either this copy or some newer copy
+              allCopies.find(existingCopyInfo => existingCopyInfo.copyNumber >= copyInfo.copyNumber)
+            case None =>
+              None
```
rjmac, line 390: I rewrote it into a match because it felt weird having the takeRollupLocks side-effects inside flatMap.

dmitry-at-tyler (approved):

PR #593, opened 2023-08-09
Remove reference to unused collocation groups

zachlefevre:


rlvoyer (approved):
Looks fine to me. I assume you verified that the config object in Scala has no reference to this parameter?

zachlefevre:
@rlvoyer there exists another PR which makes the param optional.

PR #597, opened 2023-08-25
On the new query path, do less de/recompression of responses

rjmac:
Instead of decompressing them and then re-compressing them at query-coordinator and soda-fountain, just compress them once here and then pass that through.  This means no more transparent gzip compression from jetty, so disable that for the (new) content-type that this response has.

### DIFF
--- a/soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
+++ b/soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
@@ -2,8 +2,11 @@ package com.socrata.pg.server.analyzer2
 
 import scala.collection.{mutable => scm}
 
+import java.io.{BufferedWriter, OutputStreamWriter}
+import java.nio.charset.StandardCharsets
 import java.sql.{Connection, PreparedStatement}
 import java.util.Base64
+import java.util.zip.GZIPOutputStream
 
 import com.rojoma.json.v3.ast.{JArray, JString, JValue, JNull}
 import com.rojoma.json.v3.interpolation._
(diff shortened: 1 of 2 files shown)


soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
```diff
@@ -252,32 +255,38 @@ object ProcessQuery {
...
-      for(debugFields <- debugFields) {
-        writer.write("\"debug\":%s\n ,".format(JsonUtil.renderJson(debugFields, pretty = false)))
-      }
-      writer.write("\"fingerprint\":%s\n ,".format(JString(Base64.getUrlEncoder.withoutPadding.encodeToString(etag.asBytes))))
-      writer.write("\"last_modified\":%s\n ,".format(JString(CJSONWriter.dateTimeFormat.print(lastModified))))
-      writer.write("\"locale\":%s\n ,".format(JString(locale)))
-      writer.write("\"schema\":")
+    ETag(etag) ~> LastModified(lastModified) ~> ContentType("application/x-socrata-gzipped-cjson") ~> Stream { rawOutputStream =>
+      for {
+        gzos <- managed(new GZIPOutputStream(rawOutputStream))
```
zachlefevre, line 260: ah, this is needed since it is no longer transparent.

soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
```diff
@@ -252,32 +255,38 @@ object ProcessQuery {
...
-    ETag(etag) ~> LastModified(lastModified) ~> Write("application/json; content-type=utf-8") { writer =>
-      writer.write("[{")
-      for(debugFields <- debugFields) {
-        writer.write("\"debug\":%s\n ,".format(JsonUtil.renderJson(debugFields, pretty = false)))
-      }
-      writer.write("\"fingerprint\":%s\n ,".format(JString(Base64.getUrlEncoder.withoutPadding.encodeToString(etag.asBytes))))
-      writer.write("\"last_modified\":%s\n ,".format(JString(CJSONWriter.dateTimeFormat.print(lastModified))))
-      writer.write("\"locale\":%s\n ,".format(JString(locale)))
-      writer.write("\"schema\":")
+    ETag(etag) ~> LastModified(lastModified) ~> ContentType("application/x-socrata-gzipped-cjson") ~> Stream { rawOutputStream =>
```
zachlefevre, line 258: and this is the signal throughout the stack that transparent decompression/recompression should not happen.

zachlefevre (approved):

PR #598, opened 2023-08-31
EN-62111: type-annotate number literals in the old sqlizer

rjmac:
When the old sqlizer would turn a number literal into a SQL number literal, it didn't annotate it as any particular type.  As a result, if your number literal happened to stringify as an integer, further math operations on it would be integer math operations.  As a result, for example, `select 1/4` would return 0 instead of 0.25 as would be expected.

Now instead of sqlizing as something like `1/4` that expression will sqlize as something like `(1::numeric)/(4::numeric)`.

New-sqlizer already does this


soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/Hashable.scala
```diff
@@ -124,6 +124,7 @@ object Hashable {
           hasher.hashByte(9)
           hasher.hash(lim)
           hasher.hash(off)
+        case rewrite.Pass.RemoveOrderBy => hasher.hashByte(10)
```
saraid, line 127: Feels like this is from a different working branch.

rjmac: (I added it because I'm worried about jenkins accidentally pulling it in and failing the buld because this project has error-on-warn turned on)

saraid: ah

saraid (approved):

PR #601, opened 2023-09-18
Implement SOQL => redshift SQL translation

zachlefevre:


zachlefevre:
Work done in another PR.

PR #602, opened 2023-09-25
[EN-62055]-Dalia/redshift soql

dnabilali:
added a few tests for soql -> redshift sql conversion


common-pg/src/main/scala/com/socrata/pg/analyzer2/FuncallSqlizer.scala
```diff
@@ -147,13 +163,45 @@ abstract class FuncallSqlizer[MT <: MetaTypes] extends SqlizerUniverse[MT] {
     }
   }
 
-  protected def sqlizeNormalOrdinaryFuncall(sqlFunctionName: String, prefixArgs: Seq[Doc] = Nil, suffixArgs: Seq[Doc] = Nil) = {
+  def sqlizeNormalOrdinaryFuncall(
+    sqlFunctionName: String,
+    prefixArgs: Seq[Doc] = Nil,
+    suffixArgs: Seq[Doc] = Nil,
+    castType: (Doc, Int) => Option[Doc] = (_, _) => None
```
rjmac, line 170: I'm not entirely sure I like this as-is - in particular, I think I'd prefer that if it needs to take that `Doc` parameter at all (from a quick glance I don't see anything that uses it) that it'd take an `ExprSql` and return an `ExprSql` instead of operating on `Doc`s.

common-pg/src/main/scala/com/socrata/pg/analyzer2/FuncallSqlizer.scala
```diff
@@ -147,13 +163,45 @@ abstract class FuncallSqlizer[MT <: MetaTypes] extends SqlizerUniverse[MT] {
...
+  ) = {
     val funcName = Doc(sqlFunctionName)
     ofs { (e, args, ctx) =>
       assert(args.length >= e.function.minArity)
       assert(e.function.allParameters.startsWith(args.map(_.typ)))
 
-      val sql = (prefixArgs ++ args.map(_.compressed.sql) ++ suffixArgs).funcall(funcName)
+      val castArgs = args.map(_.compressed.sql).zipWithIndex.map { case (arg, idx) =>
+        castType(arg, idx).map { cast =>
+          arg +#+ Doc("::") +#+ cast
```
rjmac, line 179: `arg.parenthesized` (it's convention in this code that parens are added at the place where there's an operator)

common-pg/src/main/scala/com/socrata/pg/analyzer2/FuncallSqlizer.scala
```diff
@@ -147,13 +163,45 @@ abstract class FuncallSqlizer[MT <: MetaTypes] extends SqlizerUniverse[MT] {
...
+
+      val argsSql = (prefixArgs ++ castArgs ++ suffixArgs)
+
+      val sql = argsSql.funcall(funcName)
+
+      ExprSql(sql.group, e)
+    }
+  }
+
+  def binaryOpCallFunctionPrefix(
```
rjmac, line 191: This is a weird function, though it's only weird because SQL is weird. :)  - I think it belongs in the SoQLFunctionSqlizerRedshift class (because in the PG class I'd insist on any special-syntax function get wrapped up in a stored proc.  I'm not super fond of the name but I don't really have any better suggestion for it off the top of my head.

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerRedshift.scala
```diff
@@ -0,0 +1,729 @@
...
+  }
+
+
+  def wrap(e: Expr, exprSql: ExprSql, wrapper: String, additionalWrapperArgs: Doc*) =
+    ExprSql((exprSql.compressed.sql +: additionalWrapperArgs).funcall(Doc(wrapper)), e)
+
+  def comment(sqlizer: OrdinaryFunctionSqlizer, comment: String) = ofs { (f, args, ctx) =>
+    val e = sqlizer(f, args, ctx)
+    ExprSql((d"/*" +#+ Doc(comment) +#+ Doc("*/") +#+ e.compressed.sql).parenthesized, f)
+  }
```
rjmac, line 34: Use an annotation for this (add a new subclass to SqlizeAnnotation, and add it to a Doc with `.ann(YourNewAnnotation(whatever))` and then teach the pretty sql renderer over in ProcessQuery about it.

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerRedshift.scala
```diff
@@ -0,0 +1,729 @@
...
+      ofs { (f, runtimeArgs, ctx) =>
+        assert(args.length >= f.function.minArity)
+        assert(f.function.allParameters.startsWith(runtimeArgs.map(_.typ)))
+
+        val expandedRuntimeArgs = args.map(_.asInstanceOf[Int]).map(n => runtimeArgs.toList(n))
+
+      val sql = Doc(sc.s(expandedRuntimeArgs.map(_.compressed.sql): _*))
+      ExprSql(sql, f)
+    }
+  }
```
rjmac, line 25: This is clever :)

rjmac (changes requested):

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerRedshift.scala
```diff
@@ -0,0 +1,729 @@
...
+
+  implicit class ExprInterpolator(sc: StringContext) {
+    def expr(args: Any*): OrdinaryFunctionSqlizer =
+      ofs { (f, runtimeArgs, ctx) =>
+        assert(args.length >= f.function.minArity)
+        assert(f.function.allParameters.startsWith(runtimeArgs.map(_.typ)))
+
+        val expandedRuntimeArgs = args.map(_.asInstanceOf[Int]).map(n => runtimeArgs.toList(n))
+
+      val sql = Doc(sc.s(expandedRuntimeArgs.map(_.compressed.sql): _*))
```
rjmac, line 22: ..though I'm not sure I like this line here.  It smashes all the Docs together into a single String and loses all annotation information.

rjmac: I think I'd prefer
```scala
  implicit class ExprInterpolator(sc: StringContext) {
    def expr(args: Int*): OrdinaryFunctionSqlizer =
      ofs { (f, runtimeArgs, ctx) =>
        assert(args.length >= f.function.minArity)
        assert(f.function.allParameters.startsWith(runtimeArgs.map(_.typ)))

        val args = args.map(runtimeArgs(_).compressed.sql)
        val sql = (sc.parts.head +: (args, sc.parts.tail).zipped.map { (arg, lit) => arg ++ Doc(lit) }).hcat
        ExprSql(sql, f)
      }
    }
```

(two things here: `expr` now takes a variable number of Ints rather than Anys, and `sc.parts` will always have one more element than `args` so we can do that head/tail thing and then hcat it all together)

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerRedshift.scala
zachlefevre: An Annotation would mean this *does not* show up in the generated SQL, but instead the annotated generated SQL when including debug info, right?

common-pg/src/main/scala/com/socrata/pg/analyzer2/FuncallSqlizer.scala
zachlefevre: Ah, good point. Doc does not make sense here. I'll remove it since it's not used.

common-pg/src/main/scala/com/socrata/pg/analyzer2/FuncallSqlizer.scala
zachlefevre: Do you then mean `(arg +#+ Doc("::") +#+ cast).parenthesized`?

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerRedshift.scala
zachlefevre: thx :)

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerRedshift.scala
zachlefevre: what is hcat? This code does not compile for me.

zachlefevre: https://hackage.haskell.org/package/pretty-1.1.3.6/docs/Text-PrettyPrint.html#v:hcat

zachlefevre: ah, the compiler is confused. I think you want Doc(sc.parts.head)

zachlefevre: @rjmac Do we want to parenthesize the `Doc(lit)` ?

common-pg/src/main/scala/com/socrata/pg/analyzer2/FuncallSqlizer.scala
rjmac: No, I meant what I said

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerRedshift.scala
rjmac: Right

zachlefevre: The idea was to keep the generated-SQL greppable and so to have the commentshow up in the generated SQL.
Happy to remove it if you think it's messy.

rjmac (approved):

PR #603, opened 2023-09-27
New-analyzer rollups

rjmac:


common-pg/src/main/scala/com/socrata/pg/analyzer2/CostEstimator.scala
```diff
@@ -21,7 +22,7 @@ object CostEstimator {
         throw new Exception("EXPLAIN should have returned a single result?");
       }
 
-      val result = JsonReader.fromString(rs.getString(0)).dyn(0).Plan("Total Cost").! match {
+      val result = JsonReader.fromString(rs.getString(1)).dyn(0).Plan("Total Cost").! match {
```
saraid, line 25: ...why on earth did _this_ change when the sql didn't?

rjmac: Because it was wrong before.  Result indices start from 1, not 0.

common-pg/src/main/scala/com/socrata/pg/store/RollupAnalyzer.scala
```diff
@@ -0,0 +1,215 @@
...
+      case Left(err) =>
+        logger.info("Failed to refind tables? " + err)
+        return None
+    }
+
+    val baseAnalysis = analyzer2(foundTables, userParameters.toUserParameters) match {
+      case Right(analysis) =>
+        analysis
+      case Left(err) => {
+        logger.info("Failed to analyzie?  Skipping rebuild: " + err)
```
saraid, line 194:
```suggestion
        logger.info("Failed to analyze?  Skipping rebuild: " + err)
```
🤷 greppability

saraid (approved):

PR #604, opened 2023-09-28
Make rollup-matching accept WHERE in aggregates in some cases

rjmac:


saraid (approved):
Didn't poke around as rigorously as I should have, but it mostly looks fine, so it should be fine.

PR #605, opened 2023-09-29
Aggregates are aggregates, not normal functions

rjmac:


saraid (approved):
I should have caught that. =/

PR #606, opened 2023-10-02
Only create FTS indexes on rollup if they look like materialized views

rjmac:
One of the things you can do in this new system is say "create a rollup with the query of this saved view" and while this doesn't imply automatically regenerating this rollup when that saved view changes, the fact that it was initially created from a view is propagated down to the place it's turned into a SQL table.  So, since FTS indexes are easily the slowest kind of index to create, let's only do that if it looks like the "rollup" is actually a "materialized view", since otherwise it's extremely unlikely to ever actually be used.

### DIFF
--- a/common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
+++ b/common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
@@ -24,7 +24,7 @@ import com.socrata.pg.soql.SqlizerContext.SqlizerContext
 import com.socrata.pg.store.index.SoQLIndexableRep
 import com.socrata.soql.{AnalysisContext, BinaryTree, Compound, Leaf, ParameterSpec, SoQLAnalysis, SoQLAnalyzer}
 import com.socrata.soql.analyzer.SoQLAnalyzerHelper
-import com.socrata.soql.analyzer2.{UnparsedFoundTables, FoundTablesTableFinder, DatabaseColumnName, DatabaseTableName}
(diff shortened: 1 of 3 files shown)


common-pg/src/main/scala/com/socrata/pg/store/RollupManager.scala
```diff
@@ -192,25 +192,36 @@ class RollupManager(pgu: PGSecondaryUniverse[SoQLType, SoQLValue], copyInfo: Cop
...
               d")" +#+
               sql
 
-            val searchIdxSql = new SoQLRewriteSearch[DatabaseNamesMetaTypes](searchBeforeQuery = true).
-              searchTerm(schema).
-              map { searchTerm =>
-                d"CREATE INDEX IF NOT EXISTS idx_" ++ Doc(rollupInfo.tableName) ++ d"_fts ON" +#+ Doc(rollupInfo.tableName) +#+ d"USING GIN (to_tsvector('english'," ++ searchTerm ++ d"))" ++ Doc(tablespaceSql)
+            val searchIdxSql =
+              foundTables.initialQuery match {
+                case FoundTables.Saved(_) =>
```
rlvoyer, line 213: Can you explain why `FoundTables.Saved(_)` is indicative of a materialized view?

rjmac: Because that means "this was created as 'the same query as a saved view'" - i.e., we're materializing the output of that view.

PR #608, opened 2023-10-02
EN-62907: Make analyzer1 way more willing to use table aliases

rjmac:


common-pg/src/test/scala/com/socrata/pg/soql/SqlizerJoinTest.scala
```diff
@@ -258,7 +258,7 @@ class SqlizerJoinTest  extends SqlizerTest {
 
   test("@this alias does not spread into join more") {
     val soql = "SELECT primary_type |> SELECT @c.primary_type, @x.name FROM @this as c JOIN (SELECT name, @dog.breed, @bird.bird, @f.fish FROM @dog JOIN @bird on breed=@bird.breed JOIN @fish as f on breed=@f.breed) as x ON primary_type=@x.name"
-    val expected = """SELECT "_c"."primary_type","_x"."name" FROM (SELECT "t1".primary_type_7 as "primary_type" FROM t1) as "_c" JOIN (SELECT "name_55" as "name","_dog"."breed_56" as "breed","_bird"."bird_68" as "bird","_f"."fish_68" as "fish" FROM t12 JOIN t13 ON ("breed_56" = "_bird"."breed_66")  JOIN t14 as "_f" ON ("breed_56" = "_f"."breed_66")) as "_x" ON ("_c"."primary_type" = "_x"."name")"""
+    val expected = """SELECT "_c"."primary_type","_x"."name" FROM (SELECT "t1".primary_type_7 as "primary_type" FROM t1) as "_c" JOIN (SELECT "name_55" as "name","_dog"."breed_56" as "breed","_bird"."bird_68" as "bird","_f"."fish_68" as "fish" FROM t12 JOIN t13 as "_bird" ON ("breed_56" = "_bird"."breed_66")  JOIN t14 as "_f" ON ("breed_56" = "_f"."breed_66")) as "_x" ON ("_c"."primary_type" = "_x"."name")"""
```
rjmac, line 261: This is my favorite thing ever; it's basically testing that this bug _does_ happen.

catstavi:
Okay so I've been going through this, but without more basis for how soql stuff gets composed I'm not sure the quality of my review right now. What I see changing is:

* joinTableAliases will now include joins where the uses didn't specify an alias, and use the "name" in the spot the alias would go
* when building repFrom:
    * we will now look for joinReps using alias if its present, otherwise falling back to name; previously it only used name
    * never include alias or name information in QualifiedUserColumnId; previously if there was an alias, we'd pass it in
* when building repJoins:
    * still looking up joinRep by name (never alias)
    * when creating QualifiedUserColumnId, pass in a name if there's no alias


^ I don't understand the meaning/impact of these changes D: 

And finally, the bit that did make sense to me: the "tn" part of the query will *always* have a qualifier, even if there was no alias.

common-pg/src/test/scala/com/socrata/pg/soql/SqlizerJoinTest.scala
catstavi: 😆

rjmac:
> Okay so I've been going through this, but without more basis for how soql stuff gets composed I'm not sure the quality of my review right now. What I see changing is:
> 
>     * joinTableAliases will now include joins where the uses didn't specify an alias, and use the "name" in the spot the alias would go

Yes, this is definitely correct.

>     * when building repFrom:
>       * we will now look for joinReps using alias if its present, otherwise falling back to name; previously it only used name
> 
>     * when building repJoins:
>       * still looking up joinRep by name (never alias)

These two do seem odd, don't they?  I'll have to dig at this a bit.  You're right this seems inconsistent.

>       * never include alias or name information in QualifiedUserColumnId; previously if there was an alias, we'd pass it in
>       * when creating QualifiedUserColumnId, pass in a name if there's no alias

Both these things should probably be like the latter.

rjmac:
Ok, gone through it a bit and seeing what impact changing that has on the generated sql, I'm fairly sure the code's correct as-is.

rjmac:
> So the take on the changes above are like "hard to trace all the threads for why they're different, but per outcome they are correct" ? Which is fine, just looking for a few casual tips to more insight about what all this is if possible.

Yep; unfortunately this code is very confusing.  Joins being Bad And Hard To Understand is like 90% of the reason for thew new-analyzer work.

PR #619, opened 2023-10-12
[EN-63000] - Dalia/aggregate functions

dnabilali:
the following functions have been tested for compatibility in redshift: max, min, count(*), count, count_distinct, sum, avg, median

median discrete actually is computing the discrete median for non-numeric data that can be ordered. in redshift discrete median can’t be calculated for non-numeric data.

the rest of the aggregate functions since redshift doesn’t support custom aggregates [Creating user-defined functions - Amazon Redshift](https://docs.aws.amazon.com/redshift/latest/dg/user-defined-functions.html)


common-pg/src/test/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerTestRedshift.scala
```diff
@@ -548,6 +548,43 @@ class SoQLFunctionSqlizerTestRedshift extends FunSuite with Matchers with Sqlize
...
+  test("max works") {
+    analyzeStatement("SELECT max(num)") should equal("""SELECT max(x1.num) AS i1 FROM table1 AS x1""")
+  }
+
+  test("min works") {
+    analyzeStatement("SELECT min(num)") should equal("""SELECT min(x1.num) AS i1 FROM table1 AS x1""")
+  }
+
+  test("count(*) works") {
+    analyzeStatement("SELECT count(*)") should equal("""SELECT (count(*)) :: decimal(30, 7) AS i1 FROM table1 AS x1""")
```
rlvoyer, line 561: Q: is the `decimal(30, 7)` translation here a Redshift thing? I would expect the count function to return an integer? Or is this for parity w/ Postgres?

rlvoyer (approved):
Looks good to me. Left a question... Can you explain the `decimal(30, 7)` in these SoQL -> SQL translations?

common-pg/src/test/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerTestRedshift.scala
dnabilali: as you can see here we defined the numeric data type in redshift: https://github.com/socrata-platform/soql-postgres-adapter/blob/331bdeea4127b0b27bb34f0580ef77b79162ca14/common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLFunctionSqlizerRedshift.scala#L763. in psql numeric representation is working differently, so the numericType for psql is numeric. we discussed this as a team for a while to see which numeric representation would work best for redshift.

PR #625, opened 2023-10-20
Add make_point

rjmac:
Plus ScopedResourceName moved in soql-reference


common-pg/src/main/scala/com/socrata/pg/soql/SqlFunctionsGeometry.scala
```diff
@@ -25,6 +25,7 @@ trait SqlFunctionsGeometry {
     TextToPolygon -> formatCall("ST_GeomFromText(%s, 4326)") _,
     TextToMultiPolygon -> formatCall("ST_GeomFromText(%s, 4326)") _,
 
+    MakePoint -> formatCall("ST_Point(%s, %s, 4326)", paramPosition = Some(Seq(1, 0))) _,
```
saraid, line 28: Not a fan of the magic number (esp. since it's not really a number), but we're clearly using it like this elsewhere, so whatever.

saraid (approved):

PR #632, opened 2023-10-26
Update soql-reference and make it possible to fail sqlization gracefully

rjmac:


soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
```diff
@@ -144,7 +144,10 @@ object ProcessQuery {
     val onlyOne = nameAnalyses.lengthCompare(1) == 0
     val sqlized = nameAnalyses.map { nameAnalysis =>
       val sqlizer = ActualSqlizer.choose(sqlizerType)
-      val Sqlizer.Result(sql, extractor, SoQLExtraContext.Result(nonliteralSystemContextLookupFound, now)) = sqlizer(nameAnalysis._1, new SoQLExtraContext(systemContext, cryptProviders, RewriteSubcolumns[InputMetaTypes](request.locationSubcolumns, copyCache), SqlUtils.escapeString(pgu.conn, _)))
+      val Sqlizer.Result(sql, extractor, SoQLExtraContext.Result(nonliteralSystemContextLookupFound, now)) = sqlizer(nameAnalysis._1, new SoQLExtraContext(systemContext, cryptProviders, RewriteSubcolumns[InputMetaTypes](request.locationSubcolumns, copyCache), SqlUtils.escapeString(pgu.conn, _))) match {
+        case Right(result) => result
+        case Left(nothing) => nothing
```
saraid, line 149: Huh. That feels weird.

saraid (approved):
Looks fine if that one thing I flagged is intended.

soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
rjmac: Yep.  Scala wants Left to be handled, but it's Left[Nothing] (which can never be instantiated, there are no values of type `Nothing`) and it's a subclass of every type so if it _did_ somehow produce a `Nothing` it unifies fine with `result`.  If the SqlizationError type ever changes, this will become a type error when the unification produces something useless like `Product with Serializable`.

PR #636, opened 2023-11-10
Type-annotate NULL literals in old-analyzer SQL generation

rjmac:
This also attempts to fix a related bug whereby compound columns would produce single-column null literals, but for some reason it does not? This definitely produces a multi-column ParametricSql when appropriate, but sqlization of `select null::url` still produces `SELECT (null :: TEXT)` and then tries to read two columns out of the result.  I'm not sure why!


rlvoyer (approved):
LGTM

PR #645, opened 2023-12-01
A little post-redshift-extraction cleanup

rjmac:
Just removing the explicit "Postgres" from things that had had it added.


rlvoyer (approved):
Nice. Looks good to me.

PR #647, opened 2023-12-06
Fix copies across column naming convention

rjmac:


rlvoyer (approved):
👍 Are you confident that these are the only secondary event handlers that require the change?

rjmac:
Yes, I think so.  This is a special case of having two tables involved.  I can't think of anything else that does so.

PR #651, opened 2023-12-28
Pass column hints from the input analysis to the output cjson

rjmac:
Also use them when computing the etag

### DIFF
--- a/soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
+++ b/soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
@@ -115,6 +115,13 @@ object ProcessQuery {
           }
       }
 
+    // Before doing rollups, extract our final hints (rollups can
+    // destroy hints).
+    val hintsByName = metadataAnalysis.statement.schema.valuesIterator.flatMap { schemaEnt =>
+      schemaEnt.hint.map(schemaEnt.name -> _)
+    }.toMap
+    log.debug("Result column hints: {}", Lazy(JsonUtil.renderJson(hintsByName, pretty = true)))
+
     // Intermixed rewrites and rollups: rollups are attemted before
     // each group of rewrite passes and after all rewrite passes have
     // happened.
@@ -164,6 +171,7 @@ object ProcessQuery {
     // Our etag will be the hash of the inputs that affect the result of the query
     val etag = ETagify(
       analysis.statement.schema.valuesIterator.map(_.name).toSeq,
+      hintsByName,
       stringifiedAnalysis,
       copyCache.orderedVersions,
       systemContext,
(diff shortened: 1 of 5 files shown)


soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
```diff
@@ -410,10 +420,10 @@ object ProcessQuery {
         writer.write("\"schema\":")
 
         @AutomaticJsonEncode
-        case class Field(c: ColumnName, t: SoQLType)
+        case class Field(c: ColumnName, t: SoQLType, h: Option[JValue])
 
-        val reformattedSchema = nameAnalysis.statement.schema.values.map { case Statement.SchemaEntry(cn, typ, _isSynthetic) =>
-          Field(cn, typ)
+        val reformattedSchema = nameAnalysis.statement.schema.values.map { case Statement.SchemaEntry(cn, typ, _hint, _isSynthetic) =>
+          Field(cn, typ, hintsByName.get(cn))
```
saraid, line 426: Out of curiosity, what's the benefit of pulling the hints out into hintsByName and then accessing them that way, rather than using the parameter here? Is it just for ETag stuff?

soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
```diff
@@ -164,6 +171,7 @@ object ProcessQuery {
     // Our etag will be the hash of the inputs that affect the result of the query
     val etag = ETagify(
       analysis.statement.schema.valuesIterator.map(_.name).toSeq,
+      hintsByName,
```
saraid, line 174: Am surprised you're making hints affect the ETag. I don't expect they'll matter often, but I had assumed you weren't going to let them change the results.

saraid (approved):

soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
rjmac: It's because the analysis is post-rollups, which may/will have destroyed any hints.

soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
rjmac: The etag is basically a hash of the result (implemented here as a hash of all the things that can affect the result) - since the hints affect the result, they go in the big etag mixer.

soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
rjmac: Added a comment to that effect here.

PR #659, opened 2024-01-23
Return the data-versions of all referenced datasets in the cjson header

rjmac:
Idly thinking about how to provide the functionality provided by the x-soda2-data-out-of-date header without unconditionally involving truth on the read path, and plumbing this information out would definitely be part of any such thing.


rjmac:
The big change in ProcessQuery is mostly whitespace.

saraid (approved):
Am utterly failing to spot any real differences in ProcessQuery beyond whitespace changes.

soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ProcessQuery.scala
```diff
@@ -407,42 +408,46 @@ object ProcessQuery {
...
+      LastModified(lastModified) ~>
+      ContentType("application/x-socrata-gzipped-cjson") ~>
+      Stream { rawOutputStream =>
+        for {
+          gzos <- managed(new GZIPOutputStream(rawOutputStream))
+          osw <- managed(new OutputStreamWriter(gzos, StandardCharsets.UTF_8))
+          writer <- managed(new BufferedWriter(osw))
+        } {
+          writer.write("[{")
+          writer.write("\"dataVersions\":%s\n ,".format(JsonUtil.renderJson(copyCache.asMap.mapValues(_._1.dataVersion).toSeq.sortBy(_._1), pretty = false)))
```
rjmac, line 421: This line is the difference; it's new.

PR #663, opened 2024-02-16
Cross-seconary stable sorting for provenances

rjmac:


saraid (approved):
`Thing.Rollup(foo)` is prettier than `Thing(foo, isRollup=true)` anyways.

PR #674, opened 2024-04-25
Bump soql-reference

rjmac:
And update the places we were generating column bases to generate column names instead


saraid (approved):
Yup. Meets expectations.

PR #676, opened 2024-05-01
EN-66367: Make datasets' resource names required and unique

rjmac:


soql-server-pg/src/test/resources/fixtures/select-curated-region-test-invalid-geometry.json
```diff
@@ -1,2 +1,2 @@
-[{"name":"Chili", "test_result": "Self-intersection[1 1]" }
```
rjmac, line 1: I have no idea why this changed!  But I ran the query that generates it over in the SoQLGeomFunctionTest and indeed it returns 2 1, not 1 1?  So was, somehow, the resource name's optionalness masking a bug before?

rjmac: ...and on jenkins it returns the original value??? I am so confused.

common-pg/src/main/resources/com/socrata/pg/store/schema/20240501-make-resource-name-required-and-unique.xml
```diff
@@ -0,0 +1,15 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <changeSet author="setzer" id="20240501-make-resource-name-required-and-unique" runOnChange="true">
```
saraid, line 5: setzer?

common-pg/src/main/scala/com/socrata/pg/query/QueryServerHelper.scala
```diff
@@ -252,7 +252,7 @@ object QueryServerHelper {
   val typeReps: Map[SoQLType, SqlColumnRep[SoQLType, SoQLValue]] = {
     // build fake columnInfo where most properties values do not matter except type
     val cid = new ColumnId(0)
-    val datasetInfo = DatasetInfo(DatasetId.Invalid, 0, 0, "", Array.empty[Byte], None)(null)
+    val datasetInfo = DatasetInfo(DatasetId.Invalid, 0, 0, "", Array.empty[Byte], DatasetResourceName(""))(null)
```
saraid, line 255: Am instinctively kinda heebie-jeebie about using an empty string here, but eh.

saraid (approved):

common-pg/src/main/resources/com/socrata/pg/store/schema/20240501-make-resource-name-required-and-unique.xml
rjmac: oops.  copypasta :)

common-pg/src/main/scala/com/socrata/pg/query/QueryServerHelper.scala
rjmac: Yeah, I'm not sure what it thinks it's doing here (it's definitely not _creating_ anything in the database, the queryserver doesn't have rights to do that) but it's OA anyway..

PR #677, opened 2024-05-03
EN-66368: v3 rollups use either internal names or resource names

rjmac:
Just temporarily until we update exsting new-analyzer rollups, then all rollups will use resource names and the TableName class can be removed.


rlvoyer (approved):
Seems pretty straightforward to me.

PR #682, opened 2024-05-15
Make the "resource name column is unique and not-null" migration pre-…

rjmac:
…runnable

By having it check to see if one of the constraints it creates already exists, and if so just mark the migration as having already run.


saraid (approved):
Surprised this wasn't already the case.

rjmac:
> Surprised this wasn't already the case.

Would've been if I'd thought of it!

PR #688, opened 2024-06-18
Don't attempt to take locks on rollups for discarded copies

rjmac:
They're _discarded_ so we don't care about them anymore.


saraid (approved):
Seems reasonable.

PR #690, opened 2024-06-24
Don't use pg config for dynamic system context lookups

rjmac:
This was always a hack; PG's settings system was absolutely not meant for this and it's always bothered me that we were using it even beyond the bug that if you happened to use an unset config val in a session where it had been previously set, you'd get back an empty string insetad of null.

Now instead if dynamic config vals are used in a query, we create a function in pg_temp (which is specifically for storing per-transaction stuff) and call it there.


saraid (approved):
Yeah, this is way better than abusing the pg context. I didn't like that, either. And as a bonus, it might even be a strategy usable for Redshift.

rjmac:
> And as a bonus, it might even be a strategy usable for Redshift.

As far as I'm aware, redshift has no equivalent to pg_temp

PR #692, opened 2024-07-08
Use COPY to generate query results

rjmac:


saraid (approved):
I don't see any issues.

PR #694, opened 2024-07-10
Add an ENCODING parameter to the COPY command used for querying

rjmac:
Shouldn't actually do anything right now, but it keeps us robust in the face of configuration changes - now both sides of this read explicitly expect to be using UTF-8, instead of one side being implicit.


saraid (approved):
Future-proofing good.

PR #696, opened 2024-07-12
Cache more results

rjmac:
* Remove the notion of "cacheable query"
* Instead, store up to 10K of response body while streaming it
* If the response didn't overflow and the full body was sent (i.e., there was no exception during processing), stick it in the cache so it can be reused


soql-server-pg/src/main/scala/com/socrata/pg/server/analyzer2/ResultCache.scala
```diff
@@ -2,13 +2,23 @@ package com.socrata.pg.server.analyzer2
...
 
 import com.socrata.http.server.util.{EntityTag, WeakEntityTag, StrongEntityTag}
-import com.socrata.soql.analyzer2._
+
+final abstract class ResultCache
 
 object ResultCache {
+  val log = LoggerFactory.getLogger(classOf[ResultCache])
+
+  private val uselessCache = new AtomicLong(0)
```
rlvoyer, line 20: What's the `uselessCache` for?

rjmac: Leftover; already deleted.

saraid (approved):
LGTM. I wonder if the limit should be configurable instead of hardcoded?

saraid (approved):
Not spotting any problems.

PR #705, opened 2024-10-24
EN-69912: Prevent the old-analyzer from generating bad aliases

rjmac:
PG column names can be of any length, but only the first 63 characters are significant.  This is a problem if two SoQL aliases exist which are longer than that.  So, we'll remap such columns at sql-generation time to i#### (where #### is just an incrementing counter as they're encountered).

This is tricker than it sounds, because it's entirely possible for a user-specified column to be called "i####" and we don't want to collide. So for those columns, we'll remap even if the name is below the pg length limits.

Why aren't we just remapping everything?  Because we have no visibility into which columns are physical database columns (and so must not be renamed) and which columns are intermediate (and so can be).  Also there's a massive pile of tests that assume column names don't ordinarily get messed with.


saraid (approved):
Well, I'm not seeing any problems.

PR #719, opened 2025-02-11
Make it possible to cache coarsended expressions of get_utc_date

rjmac:
Before:
* calling `get_utc_date` would make your query's `Last-Modified` and `ETag` headers change every second

Now:
* `Last-Modified` still changes every second, but the `ETag` (and hence also the automatic response cache) will take into account certain "coarsening" function calls surrounding the `get_utc_date` call.  In particular, if you _immediately_ convert the result of `get_utc_date` into a `floating_timestamp` in some literal time zone and then use any of the `date_extract_...` functions or any of the field accessors that return values at day-resolution or coarser, the etag will be computed with that coarsened value instead.


saraid (approved):
Dense. I can't spot any problems, but I don't have high confidence.

PR #722, opened 2025-02-21
Create cheaper exists endpoint

zachlefevre:


soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
```diff
@@ -116,60 +117,82 @@ class QueryServer(val dsInfo: DSInfo, val caseSensitivity: CaseSensitivity, val
...
   }
 
-  object RollupResource extends SimpleResource {
-    override val get = rollups _
+  object ExistsResource extends SimpleResource {
+    override val get = exists _
+    def exists(req: HttpRequest): HttpResponse = {
+      val servReq = req.servletRequest
+
+      val dsOp = Option(servReq.getParameter("ds"))
```
rjmac, line 164: Making this optional is kinda weird but ok.

rjmac (approved):

soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
zachlefevre: oh that is weird. Fixed!

soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala
```diff
@@ -161,16 +161,13 @@ class QueryServer(val dsInfo: DSInfo, val caseSensitivity: CaseSensitivity, val
     def exists(req: HttpRequest): HttpResponse = {
       val servReq = req.servletRequest
 
-      val dsOp = Option(servReq.getParameter("ds"))
+      val ds = Option(servReq.getParameter("ds")).getOrElse(throw new IllegalArgumentException("Existence Checking requires the dataset parameter"))
```
rjmac, line 164: That'll 500.  If you want to make it required, make it not a query parameter at all - `/exists/{ds}` instead of `/exists?ds=...`

PR #725, opened 2025-03-19
Add reps for Blob and Photo

rjmac:


common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLRep.scala
```diff
@@ -872,6 +872,94 @@ abstract class SoQLRepProvider[MT <: MetaTypes with metatypes.SoQLMetaTypesExt w
         }
     },
 
+    SoQLBlob -> new SingleColumnRep(SoQLBlob, d"text") {
```
rjmac, line 875: This "text" is the SQL type backing the column

catstavi: thanks, I was getting tripped up comparing this to SoQLDocument. Though I guess I thought Document and Photo were both underlyingly blob type?

rjmac: Yeah, they're not - blob and photo are both just pointers into the core blobstore.  Document is that plus some metadata.

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLRep.scala
```diff
@@ -872,6 +872,94 @@ abstract class SoQLRepProvider[MT <: MetaTypes with metatypes.SoQLMetaTypesExt w
...
+          override def csvify(cv: CV): Seq[Option[String]] = {
+            cv match {
+              case SoQLNull => Seq(None)
+              case SoQLBlob(t) => Seq(Some(t))
+              case other => badType("blob", other)
+            }
+          }
+          override def placeholders = Seq(d"?")
+          override def indices =
+            createSimpleIndices(namespace.indexName(tableName, columnName), tableName, compressedDatabaseColumn(columnName))
```
saraid, line 916: Kinda surprised we're bothering to create indices, but I guess also why not.

saraid (approved):

common-pg/src/main/scala/com/socrata/pg/analyzer2/SoQLRep.scala
catstavi: blob and photo both potentially have additional metadata columns, though right? Like blobs can have a name and Photos can have a name and a content_type
https://cate-scgc.test-socrata.com/resource/22g9-33r8.json

catstavi: (exploring around I see that content_type and filename are attributes of Document, and the other two are simple, like Robert's saying. Cool!

takahiko-naito:
When will this change appear in RC?

PR #738, opened 2025-05-20
URL-with-token for marathon

rjmac:
ECS will just get it out of secretsmanager


soql-server-pg/docker/ship.d/run
```diff
@@ -1,5 +1,7 @@
-#!/bin/sh
+#!/bin/bash
+
 set -ev
+set -o pipefail
```
rjmac, line 4: Just good hygiene here

PR #751, opened 2025-08-19
HF-71: Also handle URL subcolumn order

rjmac:


ds2643 (approved):

ds2643:
> Ceremonial question: should the PR title be HF-71?

Yes, good call, Michael.

ds2643:
And the commit header.

PR #756, opened 2025-09-03
EN-68022: Remove a pg instance's registration when it's full

rjmac:
In an attempt to force requests to other instances rather than getting to the max queue size and then forcibly closing connections.


saraid (approved):
I spy no issues.

